PR #4, opened 2017-11-15
Update year range to 2017

ASVincent:


omad (commented):
Could we do current year instead of hard coding it?

ASVincent:
@omad I was just looking for a quick solution as I don't think this is our biggest priority right now.

PR #15, opened 2018-08-23
Completion reports

ashoka1234:
**Completion reports: comparative summaries and differences**
It's currently really tedious to compare products in cubedash and we would like to be able to do it to get a clear picture of how much of different types of products have been processed, and where we might have gaps or double ups. In particular, for each month there should be the same number of products for PQ, PQ-Legacy, and WOfS. And there should be the same number of products for NBART and FC.

This feature runs on a separate page on the client side while on the server it utilises existing infrastructure, architecture, and styles. However, some client state management is implemented with `javascript` and `jquery`. The `url` of the feature is `<cubedash-url>/reports`. The server side routing is implemented as a `flask` blueprint `reports`.

This feature refer to the item `DEADM-20` in JIRA.

Integration tests are passed.

### DIFF
--- a/cubedash/_pages.py
+++ b/cubedash/_pages.py
@@ -10,7 +10,7 @@ from werkzeug.datastructures import MultiDict
 import cubedash
 import datacube
 from datacube.scripts.dataset import build_dataset_info
-from . import _filters, _dataset, _product, _platform, _api, _model
+from . import _filters, _dataset, _product, _platform, _api, _model, _reports
 from . import _utils as utils
 from ._utils import as_json
 
@@ -20,6 +20,7 @@ app.register_blueprint(_api.bp)
 app.register_blueprint(_dataset.bp)
 app.register_blueprint(_product.bp)
 app.register_blueprint(_platform.bp)
+app.register_blueprint(_reports.bp)
 
 _LOG = structlog.getLogger()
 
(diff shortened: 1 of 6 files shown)


jeremyh:
Thanks Ashoka – I didn't realise you were working on this.

Some of the files might be uncommitted; you probably have a change to `_pages.py` to register your blueprint too?

Long-term I'd like to integrate the functionality with the main UI, but if this gets us some functionality immediately it sounds good :)

ashoka1234:
Thanks @jeremyh for the reply. Yes I must have not committed `_pages.py`.  will do

omad (commented):
This looks good to me, I'm keen to try it out...

However, I don't know `cubedash` anymore, so can you please make the call @jeremyh .

jeremyh (approved):
👍 The styling is slightly broken, but as it's a "hidden" URL I'm not too concerned for now. It gives numbers that aren't otherwise available.

jeremyh:
Ah -- @ashoka1234. Awful timing for us: the Datacube 1.6 release on the weekend broke the develop branch that you're cloned against. 

No need to fix it here: the branch `next-core` was tracking changes for Datacube 1.6. I can merge that branch in now that it's released.

ashoka1234:
That's great @jeremyh

PR #17, opened 2018-08-27
Store summaries in DB, rather than files

jeremyh:
More polish can still be done, but it now has feature-parity with the current version, the tests are _significantly_ better than before, and it's clearly the way forward now. I prefer not to keep branches open for so long that are this far-reaching across the codebase.


ashoka1234:
Summaries in DB probably makes sense and I think it makes further feature development bit easier and storage efficient.  When summaries get larger, database queries could be faster or made faster relative to file store, I guess.

PR #28, opened 2018-09-18
Add Dockerfile

roarmstrong:


jeremyh (approved):
Thanks Robbie.

I might revisit the use of gunicorn soon, particularly if we're running with only one worker.

PR #39, opened 2018-10-11
Include optional dependencies in Docker instance

jeremyh:
- They make page load faster (particularly deserialisation of large datacube metadata).
- Run using [Meinheld](https://github.com/mopemope/meinheld) instead of the default gunicorn worker.


roarmstrong (approved):
Looks good!

PR #46, opened 2019-01-17
Configurable Branding of Cubedash

ashoka1234:
# Reason for this PR
This PR attempt to implement re-branding of cubedash and make it theme-configurable. A theme-based solution was sought so that broader customisation could be implemented in addition to re-branding with a configurable logo.

( Also there is a stray `css and bug fix` for the hidden `_reports.py` blueprint. ) 

# Solutions
[ ] `flask_theme` library is used to customise `jinja` templates and static files and organise and manage those as `themes`. 
[ ] A `DEFAULT_THEME` is coded in `_model.py`. In the absence of `THEME` specified in `settings.env.py`, the `DEFAULT_THEME` is used to render templates.
[ ] A `render template` function is defined in `_utils` that identify the theme to be used and render templates. This render function is used to render all the templates at the respective endpoints.  
[ ] At this stage two themes is provided: `dea` for Digital Earth Australia theme and `odc` for Open Data Cube theme.


cubedash/_utils.py
```diff
@@ -44,6 +46,10 @@
 
 _LOG = structlog.get_logger()
 
+def render(template, **context):
+    theme = flask.current_app.config.get('THEME', flask.current_app.config['DEFAULT_THEME'])
```
jeremyh, line 50: I don't think we need both properties (we're already using [setdefault() above](https://github.com/opendatacube/dea-dashboard/pull/46/files#diff-456fe973ee5300fc13637fe6e05c6e11R34))

I lean towards replacing it with one "CUBEDASH_THEME" property.

setup.py
```diff
@@ -39,6 +39,7 @@
         'dataclasses',
         'datacube>=1.6',
         'flask',
+        'flask_themes @ git+https://git@github.com/maxcountryman/flask-themes@master',
```
jeremyh, line 42: It's a shame there's no release available. We might pin it later...

cubedash/themes/odc/templates/logo.html
```diff
@@ -0,0 +1 @@
+<img src="{{ theme_static('odc-logo-horizontal.svg') }}" height="26px" alt=""/>
```
jeremyh, line 1: I might fiddle with this later. The logo looks out of place in the current colour scheme. Perhaps we'll switch to the monochrome/white version of the logo in the short term until colour theming is available.

jeremyh (approved):
Cool, thanks Ashoka.

PR #47, opened 2019-01-21
Adjust default logo, mention theme in the docs

jeremyh:
The current cubedash colours clash with the coloured ODC logo, and cubedash
colours are not yet configurable in the theming. 

For now I think the monochrome logo is a better fit:

screen shot 2019-01-22 at 10 34 54 am

Note that the official ODC website doesn't use the colour logo, so we have previously-approved comradery.


omad (approved):
:+1:

Issue #40, opened 2018-10-31
Rename repository


omad:
We should rename this repository to remove `dea` from the name.

The obvious and lazy name would be `datacube-dashboard`.

But I don't think `dashboard` is the most helpful name. Something like `datacube-explorer` might be better.

jeremyh:
Good idea, and explorer does sound more descriptive.

We probably want to do #12 around the same time.

Issue #12, opened 2018-06-28
Remove DEA branding from vanilla install


GypsyBojangles:
To encourage other groups to use "Cubedash", we should externalize our DEA branding configuration or at least make it optional.

jeremyh:
Done by Ashoka in #46 

(further branding improvments can be done, but it at least says "ODC" now instead of "DEA")

PR #49, opened 2019-01-22
Rename proposal: dea-dashboard → datacube-explorer

jeremyh:
As suggested by @omad in #40, now that the branding has been changed (#12)

(I'm happy to modify this if we want to further discuss other name ideas)

If accepted, we'll rename the repo to match.


omad (commented):
Sounds good to me, do we want to put it out to a wider audience and risk bike shedding, or are we @jeremyh and @GypsyBojangles happy to just do it?

I keep changing my mind.

GypsyBojangles (approved):

GypsyBojangles:
I think it’s a good name!

PR #50, opened 2019-01-24
Stac support

ashoka1234:
## `STAC-API` for `datacube-explorer`
This is a minimalist implementation of `STAC-API` spec for `datacube-explorer`. It consists of a single endpoint `/stac/search`  where `HTTP POST` is a required implementation while `HTTP GET` is an equivalent optional implementation. The search fields `bbox` (bounding box) and `time` are required fields for `STAC-API` while `product` field is made required for our purposes. The endpoint returns a list of `STAC items` that matches the query parameters in a `GeoJson FeatureCollection` format. If the query has more datasets that could be sent in a single return, `paging` is supported via a `link` in the `links` field of `GeoJson FeatureCollection`. The `STAC-API` spec could be found [here](https://github.com/radiantearth/stac-spec/tree/master/api-spec).

There are no changes done to `datacube-explorer` front-end rendering aspects and therefore testing could be done directly from the browser and/or using `curl`. The `API` is implemented as a `Flask blueprint` in `_stac.py`.


jeremyh:
Thanks Ashoka.

Can you add some tests before we review it? Coveralls was broken because of the recent repository rename, so it hasn't complained here. -- I've just fixed it.

There's code that provides test data and configuration already: see the (basic) examples for testing the current geojson api and front-end pages: https://github.com/opendatacube/datacube-explorer/blob/c99aede051489d353595b8c7b53983076b0c49a9/integration_tests/test_page_loads.py#L172

ashoka1234:
Thanks Jeremy.

Sure will add the tests.

ashoka1234:
Added  single test that cover both `GET` and `POST`. If you needed further tests or more breath please let me know

jeremyh (dismissed):
Thanks Ashoka.

From running it locally, there’s a few errors coming back. It might be worth adding tests for them:

(several of these shouldn't throw server exceptions as they're user errors, and we’ll receive system notifications.)

- The “next” page link should be valid (test that the url can be loaded).
- User forgetting required fields. Ideally return an understandable error message to the user too, eg: https://github.com/opendatacube/datacube-explorer/blob/a57f72390868d298147b60d0282bcd2a389817a5/cubedash/_product.py#L18-L19
- A minimal query should work (eg. no “limit” specified by user)
- A query that returns no results (eg. out of the time/space bounds).
- A query for a product that uses geo_ref_points (I assume it's intended to be supported as there's some code for it).

(I'm excited about tests here, as they have a pretty high bang-for-the-buck: it’s a read-only api, publicly facing, with a strong specification, and there’s already test data and setup for it.)

I'm guessing the original prototype code was only written for a few specific products in the cloud instance, as it's returning blank values for most of the existing dea collection products. I know it's a minimal initial implementation, so its probably intentional, but we should make it clear if so. (even if that means returning an "unsupported product" error to the user, it's better than quiet invalid values).

Eg. The “assets” section paths are all blank in most products. ‘ls8_nbar_albers’:

screen shot 2019-02-01 at 10 56 55 am

Also:
- Is the root page (`"/"`) usually blank in stac?
- Does stac have a jsonschema or something we could throw into a test ("return document validates against schema")? That could catch many issues.
- You mentioned 'product' is our own extension, but is required, so presumably we're not stac compliant just yet. The underlying datacube `index.search` methods should work without a product specified (they will union all products that match the search terms).

ashoka1234:
Sure will address those. My bad! I missed parsing bbox!

ashoka1234:
Hi Jeremy,
I have done changes and refactoring and followings are comments about those:

1. Projection code (to `epsg:4326`)  refactored so that it works with broad number of projection strings.
2. next page link code refactored so that it works with more edge cases.
3. `product` is no longer compulsory and associated refactoring done.
4. `bbox` parsing done which I missed earlier. 
5. Added user error messages instead of relying on system error messages.
6. Added tests for the two products that are populated for the test run (`wofs_albers` and `high_tide_comp_20p`) including testing `next` link

However, there are few issues that needed unfulfilled at this stage:
1. blank asset links: For datasets in `.nc` files, `metadata_doc` returns empty band path names. We don't have a clear solution to this at this stage.
2. There is a `STAC` validator that could be used to validate the entire `STAC item` but we have to avoid using it until we have a clear direction of how we overlay a conceptual `STAC catalog` on `datacube database` .
3. The root page is meant to be the optional `danamic STAC catalog` which is a substantial work item that overlap with above issues so hope to implement it in a separate PR.
4. `STAC API` spec is defined in terms of `openAPI` so I believe there could be schema validators out there but not sure.  

Please let me know if you see further changes that you see good to have.

jeremyh (dismissed):
Thanks Ashoka, the next link loads and it’s good to see a load test. Please add tests when you fix other issues too! For instance: remove “limit” from one of the test requests, so that we know it now works without a limit. There were a few other cases I mentioned in the previous review that weren't/aren't working and aren't tested.

Note that the testing can be simpler/easier than it is now — see [the existing test I pointed to earlier](https://github.com/opendatacube/datacube-explorer/blob/d3ac35b750a2fdf4402e90554f5de29c0c18df5b/integration_tests/test_page_loads.py#L184-L208). It does almost the same thing that we want to do for stac: 
- it's hitting urls for a geojson response. 
- it's checking that the return was successful, via a helper method, which the current stac test doesn't.
- the test itself is querying the same test data, in a similar way, so the returned data counts will be the same as our stac api.

Feel free to copy it! (and improve it — stac has paging, spatial bounds, etc)

There are other tests in the same file that might be worth adapting too, eg:
https://github.com/opendatacube/datacube-explorer/blob/d3ac35b750a2fdf4402e90554f5de29c0c18df5b/integration_tests/test_page_loads.py#L209-L224

I'm still hitting a lot of bugs in the basic behaviour (simple searches+limits+paging), and it dies with exceptions on many of our products. Add some tests that use the limits and paging. For instance: does the correct total count of datasets come back across all pages? What if the limit/page is smaller/larger than results? And when some query params are omitted? There's only a small amount of test data, so we can be exact in our asserts. It's quite error-prone logic, and we've already seen many bugs.

The exceptions thrown seem to be related to the lat/lon extraction code. Rather than fix this code, it might be better to change from using `datasets.search_returning()` to `datasets.search()`. This gives us Dataset objects, which have bounds already, so we won't have to calculate them ourselves and handle the edge cases.

After reading the stac docs I think several of the fields are incorrect too. We can discuss that separately once the basic querying is working -- you mentioned already that there's planned work to map our internal structures onto stac.

ashoka1234:
Hi Jeremy,
I have added hard a limit and updated tests for testing GET, POST, links, limit, product, and assets. Also note that I have added JSON schema validator of STAC assets to tests as well. The code that handle the limit is refactored so that it works when no product is specified.

ashoka1234:
Hi Jeremy, I may have now added all the tests you wanted

cubedash/_stac.py
```diff
@@ -0,0 +1,235 @@
...
+    """
+)
+
+MAX_DATASETS = 100
+DATASETS_PER_REQUEST = 20
+
+
+@bp.route('/')
+def root():
+    return utils.as_json({})
```
jeremyh: A blank response might be confusing to users -- it looks like a mistake, and it may confuse some stac clients.

A 404 error could more explicitly tell them that a root catalog is not support yet:
```suggestion
    abort(404, "Only /stac/search is currently supported")
```
(perhaps a 501-not-implemented error is more correct, but it's cacheable as a permanent response, which doesn't sound wise to me. All other unimplemented paths are 404s by default, so I'm leaning towards it.)

cubedash/_stac.py
```diff
@@ -0,0 +1,235 @@
...
+        req_data = request.get_json()
+        bbox = req_data.get('bbox')
+        time_ = req_data.get('time')
+        product = req_data.get('product')
+        limit = req_data.get('limit')
+        from_dts = req_data.get('from')
+
+    # bbox and time are compulsory
+    if not bbox:
+        abort(404, "bbox must be specified")
```
jeremyh: Use status code 400 for bad arguments, not 404. (there's a few more in this file to change too)
```suggestion
        abort(400, "bbox must be specified")
```

See the list of different codes here: https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml
(or more friendly: https://httpstatuses.com/)

(Note that `bbox` and `time` are optional in the stac spec, but we can implement that later.)

cubedash/_stac.py
```diff
@@ -0,0 +1,235 @@
...
+    bbox = json.loads(bbox) if isinstance(bbox, str) else bbox
+    time_ = time_ if isinstance(time_, str) else json.dumps(time_)
+    limit = int(limit) if isinstance(limit, str) else limit
+    if from_dts and isinstance(from_dts, str):
+        from_dts = int(from_dts)
+
+    # from_dts must be lower than limit
+    limit = MAX_DATASETS if not limit else min(limit, MAX_DATASETS)
+
+    if from_dts and from_dts >= limit:
```
jeremyh: ?
```suggestion
    if from_dts and from_dts >= MAX_DATASETS:
```

cubedash/_stac.py
```diff
@@ -0,0 +1,235 @@
...
+    result = dict()
+    result["type"] = "FeatureCollection"
+    res_bbox = _compute_bbox([dt.metadata_doc_without_lineage() for dt in datasets_])
+    if res_bbox:
+        result["bbox"] = res_bbox
+
+    result["features"] = [stac_dataset(dt) for dt in datasets_]
+
+    if len(list(itertools.islice(datasets, 1))) == 1 and to_dts < limit:
+        url_next = host_url + 'stac/search?'
```
jeremyh: I doesn't really matter for this pull request, but we could use Flask's `url_for()` in the future to simplify this url code.

cubedash/_stac.py
```diff
@@ -0,0 +1,235 @@
...
+
+    query = dict()
+    if product:
+        query['product'] = product
+    time_period = time.split('/')
+    if len(time_period) == 2:
+        query['time'] = Range(Timestamp(time_period[0]), Timestamp(time_period[1]))
+    elif len(time_period) == 1:
+        query['time'] = Range(Timestamp(time_period[0]).floor('1D'),
+                              (Timestamp(time_period[0]) + DateOffset(1)).floor('1D'))
```
jeremyh: The `floor()`ing here doesn't look right to me. If I search using `time=2018-02-12T23:20:50Z` it will return datasets of completely different times on the same day.

I do like that a value without time ("2017-04-31") will conveniently search a whole day range though.

The stac spec doesn't clarify this case as far as I can see. Taken literally it it currently requires time:

> 
> - A date-time: "2018-02-12T23:20:50Z"
> - A period: "2018-02-12T00:00:00Z/2018-03-18T12:31:12Z" or "2018-02-12T00:00:00Z/P1M6DT12H31M12S"
> 
> Only features that have a temporal property that intersects the value of time are selected.

The example implementation from the stac repo seems to support day-only as a range: https://sat-api.developmentseed.org/stac/search?datetime=2019-02-18

But ignores seconds (& milliseconds) when matching exact times: https://sat-api.developmentseed.org/stac/search?datetime=2019-02-18T00:37:51.660000+00:00
(I'm not sure whether we should copy this behaviour)

(I'm also slightly averse to using `pandas` here because it's _very_ lenient with accepted date formats, and we'll need to continue supporting all of those other date variations in the future if people start using them. Pandas also isn't currently declared as a dependency)

cubedash/_stac.py
```diff
@@ -0,0 +1,235 @@
...
+    if not time_zone:
+        center_dt = center_dt.replace(tzinfo=datetime.timezone.utc).isoformat()
+    else:
+        center_dt = center_dt.isoformat()
+
+    # parent? We will have an empty parent for now
+    stac_item = OrderedDict([
+        ('id', metadata_doc['id']),
+        ('type', 'Feature'),
+        ('bbox', [metadata_doc['extent']['coord']['ll']['lon'],
```
jeremyh: Unfortunately we can't assume which of `ll`/`ur`/`ul`/`lr` have the highest/lowest values -- it will break on many existing datasets in our collections.

As I mentioned in the previous review, it's less error prone to use the Dataset's existing extent code (`dataset.extent`) for all of this, as it handles more edge cases already.

There's also a local utils method that handles a few other edge cases:
```
    from cubedash import _utils
    shape, is_valid_extent = _utils.dataset_shape(dataset)
    bbox = shape.bounds
```

(Note that `shape` may be null if the extent is missing or invalid and not repairable)

cubedash/_stac.py
```diff
@@ -0,0 +1,235 @@
...
+            'homepage': CFG['homepage']
+        }),
+        ('links', [
+            {'href': uri, 'rel': 'self'},
+        ]),
+        ('assets', {
+            band_name: {
+                # "type"? "GeoTIFF" or image/vnd.stac.geotiff; cloud-optimized=true
+                'href': band_data['path'],
+                "required": 'true',
```
jeremyh: is `required: true` a stac thing? I can't see it anywhere. What does it mean?

cubedash/_stac.py
```diff
@@ -0,0 +1,235 @@
...
+                "type": "GeoTIFF"
```
jeremyh: `type: GeoTIFF` will be wrong on most of our products (and the [stac spec](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#media-types) says to use `image/tiff or image/vnd.stac.geotiff`*).

A missing field is better than a wrong field. We should either sniff the file extension or remove it for now, I think.

\* the example `sat-api.developmentseed.org` seems to use a third one: `image/x.geotiff`

integration_tests/test_page_loads.py
```diff
@@ -383,3 +384,87 @@ def test_with_timings(client: FlaskClient):
...
+    dataset_count = DATASETS_PER_REQUEST
+
+    # Test the links and paging and when paging stop
+    next_links = [link['href'] for link in geojson['links'] if link['rel'] == 'next']
+    while next_links:
+        geojson = get_geojson(client, next_links[0])
+        assert len(geojson.get('features')) == DATASETS_PER_REQUEST
+        dataset_count += DATASETS_PER_REQUEST
+        next_links = [link['href'] for link in geojson.get('links', []) if link['rel'] == 'next']
+    assert dataset_count == MAX_DATASETS
```
jeremyh: Nice :+1:, testing that all pages add up!

integration_tests/test_page_loads.py
```diff
@@ -383,3 +384,87 @@ def test_with_timings(client: FlaskClient):
...
+
+    # Test POST, product, and assets
+    rv: Response = client.post('/stac/search',
+                               data=json.dumps({'product': 'wofs_albers',
+                                                'bbox': [114, -33, 153, -10],
+                                                'time': '2017-04-16T01:12:16/2017-05-10T00:24:21',
+                                                'limit': DATASETS_PER_REQUEST}),
+                               headers={'Content-Type': 'application/json', 'Accept': 'application/json'})
+    geodata = json.loads(rv.data)
+    assert len(geojson.get('features')) <= DATASETS_PER_REQUEST
```
jeremyh: Make this an exact equality. We should know how much of our test data matches the query. (Ten, I think. But the limit is 5. Perhaps raise the limit to make the test more useful?)

There's also a subtle typo: `geojson` was the variable for the previous test's response, not this one (it should be `geodata`). We could split these paragraphs into separate test methods to make this mistake less likely.

integration_tests/test_page_loads.py
```diff
@@ -383,3 +384,87 @@ def test_with_timings(client: FlaskClient):
...
+                                                'time': '2000-01-01T00:00:00/2016-10-31T00:00:00',
+                                                'limit': 5}),
+                               headers={'Content-Type': 'application/json', 'Accept': 'application/json'})
+    geodata = json.loads(rv.data)
+    bands = ['blue', 'green', 'nir', 'red', 'swir1', 'swir2']
+    for band in bands:
+        assert band in geodata['features'][0]['assets']
+
+    # Validate stac items with jsonschema
+    with open('integration_tests/schemas/stac_item.json', 'r') as fp:
```
jeremyh: This path will fail when the tests are run from a different directory.

See the geojson schema for an example that loads relative to the current file:

https://github.com/opendatacube/datacube-explorer/blob/1783ccf830d95357e4e6ad8dc5657c1e70010663/integration_tests/asserts.py#L18-L19

cubedash/_stac.py
```diff
@@ -0,0 +1,235 @@
...
+import itertools
+
+from . import _utils as utils
+from . import _model
+
+
+_LOG = logging.getLogger(__name__)
+bp = Blueprint('stac', __name__, url_prefix='/stac')
+
+CFG = yaml.load(
```
jeremyh: It doesn't have to be in this pull request, but we should move these to the settings module before release. Other users running the code shouldn't have "Geoscience Australia", and our phone number, returned as their default organisation.

cubedash/_stac.py
```diff
@@ -0,0 +1,235 @@
...
+
+    datasets = load_datasets(bbox, product, time, limit)
+
+    from_dts_ = from_dts or 0
+    to_dts = min(limit, from_dts_ + DATASETS_PER_REQUEST)
+    datasets_ = list(itertools.islice(datasets, from_dts_, to_dts))
+
+    result = dict()
+    result["type"] = "FeatureCollection"
+    res_bbox = _compute_bbox([dt.metadata_doc_without_lineage() for dt in datasets_])
```
jeremyh: We can use `dt.metadata_doc` as search results have no lineage already.

`dt.metadata_doc_without_lineage()` is a brand new method that'll give us a hard requirement on an unreleased version of datacube [and whose name/implementation I've just seen and am slightly hesitant about].

cubedash/_stac.py
```diff
@@ -0,0 +1,235 @@
...
+    if res_bbox:
+        result["bbox"] = res_bbox
+
+    result["features"] = [stac_dataset(dt) for dt in datasets_]
+
+    if len(list(itertools.islice(datasets, 1))) == 1 and to_dts < limit:
```
jeremyh: `to_dts < MAX_DATASETS`?

(aside: the purpose of the first clause could be made more clear: such as by adding an `iterator_has_next()` util method or a comment).

cubedash/_stac.py
```diff
@@ -0,0 +1,235 @@
...
+    # bbox is in GeoJSON CRS (WGS84)
+    query['lon'] = Range(bbox[0], bbox[2])
+    query['lat'] = Range(bbox[1], bbox[3])
+
+    return _model.STORE.index.datasets.search(limit=limit, **query)
+
+
+def stac_dataset(dataset):
+    """
+    Returns a dict corresponding to a stac item catalog
```
jeremyh:
```suggestion
    Returns a dict corresponding to a stac item
```

cubedash/_stac.py
```diff
@@ -0,0 +1,235 @@
...
+    """
+    metadata_doc = dataset.metadata_doc_without_lineage()
+    uri = dataset.local_uri
```
jeremyh: `dataset.local_uri` will break the current "datasets on s3" use case, as it only returns `file://` urls.

For now it may be better to grab the first uri from the `dataset.uris` array, which includes both local and remote uris (or if enthusiastic, filter for remote uris first.). 

How we handle/translate uris correctly will be a bigger future task.

cubedash/_stac.py
```diff
@@ -0,0 +1,235 @@
...
+        points = [[list(point.values()) for point in
+                   metadata_doc['grid_spatial']['projection']['geo_ref_points'].values()]]
+
+        # last point and first point should be same
+        points[0].append(points[0][0])
+
+        geodata = valid_coord_to_geojson({'type': 'Polygon', 'coordinates': points}, dataset.crs)
+
+    # Convert the date to add time zone.
+    center_dt = parse(metadata_doc['extent']['center_dt'])
```
jeremyh: `dataset.center_time` will work with more of our products, and is already a datetime.

jeremyh (dismissed):
Thanks Ashoka -- the tests are much better now! :+1:

There's only a few small things remaining.

(I've left out most comments on the stac `Item` document properties as I think we'll fix them in a future pull request.)

cubedash/_stac.py
ashoka1234: I wanted to keep `from_dts` within `limit` specified so that `paging` occur only within `limit` . The reason was that we anyway truncate at `MAX_DATASETS` . This could be seen as deviating from the spec but `MAX_DATASETS` also forces us to deviate. Not really sure about this!

cubedash/_stac.py
ashoka1234: Done

cubedash/_stac.py
ashoka1234: Wanted to keep paging within `limit` specified since anyway we are truncating at `MAX_DATASETS`

cubedash/_stac.py
ashoka1234: done

cubedash/_stac.py
ashoka1234: done

cubedash/_stac.py
ashoka1234: Agreed

cubedash/_stac.py
ashoka1234: done

cubedash/_stac.py
jeremyh: I think this is breaking from the stac spec in a way that will confuse users, and I can't see any benefits to doing it this way (it's not any simpler, right?).

If I understand it correctly:
- The stac spec behaviour is that specifying `limit=3`, I'll get 3 results on a page, and I can keep loading pages until we hit our own internal `MAX_DATASETS` limit (which we may remove in the future).
- With this behaviour, if I query with `limit=3`, I'll get 3 results and a later page will throw an error because it's higher than 3. Stac clients that already specify their own page sizes may not realise they're limiting the total datasets they can see. Worse, they will need to change it to 10000 to see all 10000 datasets up to the `MAX_DATASETS` limit (which we don't want them to start doing!).

If I really want to limit myself to 3 results total, I can already do that by setting `limit=3` and  ignoring the next page; I don't need the second behaviour to do it for me.

ashoka1234: Hmm OK I will change this. I had interpreted `limit` as total required rather than a possible page size

cubedash/_stac.py
ashoka1234: I have carried this thing over from earlier work. Either it was present in earlier STAC versions or Siva may have added it for some other reason. I have lost track of this may be I will remove it.

cubedash/_stac.py
ashoka1234: Not hard to sniff but I will remove it

cubedash/_stac.py
ashoka1234: done

integration_tests/test_page_loads.py
ashoka1234: That was sloppy. Done

integration_tests/test_page_loads.py
ashoka1234: Done

cubedash/_stac.py
ashoka1234: Done

jeremyh (changes requested):
Thanks for getting Travis sorted Ashoka.

I haven't done a line-by-line yet, but from a quick test it looks like there's a few regressions:
- Many requests are timing out for me. It looks like there's no longer a limit passed to the db query.
- the "next" link is often missing. (I assume this is because of the new dataset skipping, as the return count wont be "limit+1")

ashoka1234:
Hi Jeremy,
Yes `limit` had to be taken out from the query since we no longer enforce the product and the limit value is not correct. Also since we filter invalid datasets, the limit would not be correct either. If the removal of `limit` effect response time we could put some larger value greater than `MAX_DATASETS`. Counting is done on validated datasets. Next link will not be present if we reach `MAX_DATASETS` or if validated dataset generator is empty. It could be timeout affecting the generator or I don't seem to be able to reproduce that context.

jeremyh:
I'm doing some work to merge this stac logic with the existing geojson api, as long-term we wont need both, and so we may as well consolidate effort earlier.

ashoka1234:
Yes I also think that's a good idea.

jeremyh:
I've pushed the (big! sorry) merge of the two backends. It fixes many of the paging issues and adds more tests.

It also includes:
- simple implementations of the root `/stac`, product catalogs, and Item pages.
- many additions to the Item structure/properties.
     -  but there's still  a lot more of our metadata we could map across. We probably need to discuss this and test it on more real-world datasets.
- all search fields (`bbox`/`time`) are now optional. They needed to be for the other api.
- the search now uses Explorer's spatial table.

Two areas that still need improvement:
- error messages should be returned as json (as defined in stac). We return http error codes, but only plain-text errors.
- `eo:bands` implementation on Items/Collections. I started this, but my `eo:bands` arrays currently contain strings (which datacube uses for band identification), and stac expects them to be mapped to integers.

... but I think it's good enough to merge this pull request as an MVP. We can make further improvements in (hopefully smaller sized) future PRs.

jeremyh (approved):

PR #55, opened 2019-03-01
Basic Stac support

ashoka1234:
## `STAC-API` for `datacube-explorer`
This is a minimalist implementation of `STAC-API` spec for `datacube-explorer`.  The `STAC-API` spec could be found [here](https://github.com/radiantearth/stac-spec/tree/master/api-spec). This basic implementation has implemented the single endpoint `/stac/search`  where `HTTP POST` is a required implementation while `HTTP GET` is an equivalent optional implementation. The optional `stac` endpoint will be released in a future release. Currently The search fields `bbox` (bounding box) and `time` are required fields though these two fields are optional in the `STAC API schema` (Please refer to the API spec of the STAC spec). The `stac/search` endpoint returns a list of `STAC items` that matches the query parameters, in a `GeoJson FeatureCollection` format. `paging` is supported via a `link` in the `links` field of `GeoJson FeatureCollection` up to an implementation determined constant number of datasets.

There are no changes done to `datacube-explorer` front-end rendering aspects. The `API` is implemented as a `Flask blueprint` in `_stac.py`. The implementation rely on improved parsing of `Index.datasets.search`  function of the `datacube api` to properly interpret, for example `extents`. It further uses `cubedash._utils` to filter out datasets with invalid extents, and no `stac items` are generated for such datasets. `pytests` are added in the `integration_tests` module (The fixture `stac_settings` and the test function `test_stac_search`). 

The `GDAL` libray required updating to a recent version (2.4.0) due to parsing of certain datasets resulting in `exceptions`. However, this update appear to break `shared GDL libraries` in `travis CI`. As a temporary fix, a `linux symbolic link` needed to be created between  `libgdal.so` and `libgdal.so.1`.

This `PR` addresses the issues raised in the PR #50 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,11 @@ addons:
   services:
     - postgresql
   apt:
+      sources:
+      - sourceline: ppa:nextgis/ppa
       packages:
+          - gdal-bin
+          - gdal-data
           - libhdf5-serial-dev
           - libnetcdf-dev
           - libproj-dev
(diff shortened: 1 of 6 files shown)


jeremyh:
Closing, as these changes were merged into the original PR thread #50

PR #56, opened 2019-03-05
Option to not initialize the schema / install postgis

roarmstrong:
Add --init-database/--no-init-database flag to allow users to not attempt database initialization. Will default to True (attempt to initialize database) to ensure that the database will be updated if the schema changes or on first use.


jeremyh (approved):
Thanks Robbie

PR #61, opened 2019-06-07
Publish db update to sns topic

santoshamohan:
Publish new `db` `dump` `name` to `SNS` `topic` for automating db name update within the `yaml` config file.


jeremyh (approved):
:+1:

PR #66, opened 2019-07-24
Enable handling data that crosses the antimeridian

alexgleith:
This enables visualisation of scenes that cross the antemeridian.

It is likely to be slower, but hopefully not so much that it's an issue. And it's tested on the Fijian areas and works fine.

Would you like me to write some test code?


jeremyh:
Thanks for working on this Alex, it's sorely needed :)

A test would be good to stop me from accidentally breaking it in the future --- if you have an example dataset we could add it to the test data. Or alternatively add a standalone/unit test that calls the `footprint_wgs84` method directly.

The current Travis build will be fixed by running Black (or [make format](https://github.com/opendatacube/datacube-explorer#code-style)). I use [pre-commit](https://github.com/opendatacube/datacube-explorer#pre-commit-setup) locally to make it automatic.

alexgleith:
Hey @jeremyh, ah, I use a flake-8 plugin in code. I'll have a look at the black errors.

I'll see if I can get the tests working and run them in my dev environment.

jeremyh:
Thanks Alex, looks good. I might make the last check more crude and hardcode the expected polygon so that we don't have to rely on `_test_wrap_coordinates` being correct.

jeremyh:
I did some timeit runs and the speed hit looks quite large unfortunately :frowning_face:  it's showing almost a second (0.96) consistently on my laptop to run one `p.footprint_wgs84` call, up from 0.06 without the patch.
```
 $ python -m timeit -n 10 -u sec -s 'from integration_tests.test_model import make_overview; o=make_overview()' -- o.footprint_wrs84
10 loops, best of 3: 0.96 sec per loop
 $ git stash
 $ python -m timeit -n 10 -u sec -s 'from integration_tests.test_model import make_overview; o=make_overview()' -- o.footprint_wrs84
10 loops, best of 3: 0.0656 sec per loop
```

It's called on almost every page load (and most stac API calls), so we might want to look into optimisations before merging. (Or caching if there's no way around it?).

I _love_ doing optimisation, but probably won't have time at the moment with our reprocess! Happy to leave this open...

alexgleith:
Right, well perhaps it needs two paths, one that checks about vicinity to the antimeridian and the other that ignores it and does an optimised CRS transform.

Related, I noticed that the CRS for the 'original' data is Australian Albers, `3577`. Where does that come from? Since this will be used globally, and the data I indexed was in a UTM projection, it's odd that it would store the footprint in Albers. We can talk about it tomorrow maybe...

jeremyh:
> Related, I noticed that the CRS for the 'original' data is Australian Albers, `3577`. Where does that come from? Since this will be used globally, and the data I indexed was in a UTM projection, it's odd that it would store the footprint in Albers. We can talk about it tomorrow maybe...

Good catch ... that should be moved to a config file or something. It's here:
https://github.com/opendatacube/datacube-explorer/blob/10eff3dd761562cbaeeb5ebfb3975189c9963a45/cubedash/summary/_schema.py#L30-L31
The slightly tricky parts are:
- our area estimates currently assume it's in units of metres
- it's used in a Geometry column of the DB schema, which is static: https://github.com/opendatacube/datacube-explorer/blob/10eff3dd761562cbaeeb5ebfb3975189c9963a45/cubedash/summary/_schema.py#L135

alexgleith:
Right. world mercator is pseudo metres... not any good for measuring anything but relative areas.

jeremyh:
:+1: :+1: Huge improvement, it's now indistinguishable.

```
 $ python -m timeit -n 5 -u sec -s 'from integration_tests.test_model import _create_overview as overview; o=overview()' -- o.footprint_wrs84
5 loops, best of 3: 0.0667 sec per loop
```

Added the test as a pytest-benchmark for future reference.

jeremyh (approved):

PR #75, opened 2020-02-24
Set up an opinionated Docker development environment.

alexgleith:
A nice, tidy Docker DEV environment.

I run this with Cube in a Box so that I have a local application. It also has a thing so that you can run it with a local HTTPS cert... it may need more work to make it work for others.


jeremyh (dismissed):
:+1:

jeremyh (approved):

PR #76, opened 2020-02-24
Adding flag to force refresh of tables, incrementally for the most part.

alexgleith:
Added a flag to the CLI so that we can force-refresh the spatial table and the summary ones too.

Let me know if it's too naive, @jeremyh!


cubedash/summary/_stores.py
```diff
@@ -510,7 +511,7 @@ def search_items(
             query = query.where(DATASET_SPATIAL.c.id.in_(dataset_ids))
 
         if require_geometry:
-            query = query.where(DATASET_SPATIAL.c.footprint != None)
+            query = query.where(DATASET_SPATIAL.c.footprint is not None)
```
jeremyh, line 515: `is not None` is usually good practice in normal Python, but not in sqlalchemy queries. sqlalchemy overrides the `!=` operator but can't override the "is not" operator, so this changes the query behaviour.

(I'm curious if the tests caught this -- the build's broken for an unrelated reason)

One alternative, if you're getting this from a lint error, is to switch to sqlalchemy's [`isnot()` function](https://docs.sqlalchemy.org/en/13/core/sqlelement.html#sqlalchemy.sql.expression.ColumnElement.isnot), but in the past I just suppressed that particular lint error in our DB modules because I find the operator more readable (and it's used more often in sqlalchemy examples).

cubedash/generate.py
```diff
@@ -142,10 +154,12 @@ def cli(
     force_concurrently: bool,
     verbose: bool,
     init_database: bool,
+    force_refresh: bool
 ):
     """
     Generate summary files for the given products
     """
+    _LOG.info("Running process to refresh")
```
jeremyh, line 162: I'd change these human-friendly messages to `echo()` if we want them.

We're using `echo()` in this file for human cli output, and the [structlog](http://www.structlog.org) logger for machine-readable output (filterable with jq etc), hence the log keys like "`product.init`".

cubedash/summary/_stores.py
```diff
@@ -128,18 +128,19 @@ def refresh_product(
     ):
         our_product = self.get_product_summary(product.name)
 
+        _LOG.info(f"Refreshing if {our_product.last_refresh_age} is older than {refresh_older_than}")
```
jeremyh, line 131: I'd prefer to add `age=` as a struct-logged parameter below -- I was trying to keep UI messages out of the inner lib, as it's callable in Python code directly.

cubedash/summary/_stores.py
```diff
@@ -542,14 +543,15 @@ def get_or_update(
...
+        force_refresh: Optional[bool] = False
     ):
         """
         Get a cached summary if exists, otherwise generate one
 
         Note that generating one can be *extremely* slow.
         """
         summary = self.get(product_name, year, month, day)
-        if summary:
+        if summary and not force_refresh:
```
jeremyh, line 554: I don't think this will be enough.

The summaries are generated top-down: it summarises a product by grabbing all year summaries (generating if missing), and it summaries a year by fetching all 12 of the month summaries (generating if missing), etc.

So when you force-refresh a year, it will just grab the old, unrefreshed, month summaries to recreate it. (I assume?).

cubedash/summary/_stores.py
```diff
@@ -128,18 +128,19 @@ def refresh_product(
...
         our_product = self.get_product_summary(product.name)
 
+        _LOG.info(f"Refreshing if {our_product.last_refresh_age} is older than {refresh_older_than}")
         if (
             our_product is not None
             and our_product.last_refresh_age < refresh_older_than
         ):
             _LOG.debug(
-                "init.product.skip.too_recent",
+                "init.product.skipping.too_recent",
```
jeremyh, line 137: Most of the log codes use the third-person form, i'd like to keep them consistent unless we want to change all of them.

eg.https://github.com/opendatacube/datacube-explorer/blob/05fbeb731e6bd238c1646b9fed2934113e1b952b/cubedash/summary/_stores.py#L622
https://github.com/opendatacube/datacube-explorer/blob/05fbeb731e6bd238c1646b9fed2934113e1b952b/cubedash/summary/_stores.py#L630
https://github.com/opendatacube/datacube-explorer/blob/05fbeb731e6bd238c1646b9fed2934113e1b952b/cubedash/summary/_stores.py#L628
https://github.com/opendatacube/datacube-explorer/blob/591499d1377b41ed55079a5167caa31050f98fbd/cubedash/generate.py#L32

jeremyh (changes requested):
Thanks for looking at this! :slightly_smiling_face:

cubedash/summary/_stores.py
alexgleith: Ah, good catch. I changed all these to `is not` and then realised it broke all the queries. I didn't notice I hadn't put this one back.

cubedash/summary/_stores.py
alexgleith: ok ;-)

PR #77, opened 2020-02-24
Stac and minor tweaks

alexgleith:
Pretty simple


cubedash/_stac.py
```diff
@@ -319,11 +319,13 @@ def as_stac_item(dataset: DatasetItem):
         bbox=dataset.bbox,
         geometry=dataset.geom_geojson,
         properties={
+            "collection": dataset.product_name,
             "datetime": utc(dataset.center_time),
             **dict(_build_properties(dataset.odc_dataset.metadata)),
             "odc:product": dataset.product_name,
             "odc:processing_datetime": utc(dataset.creation_time),
             "cubedash:region_code": dataset.region_code,
+            "eo:epsg": ds.crs.epsg
```
jeremyh: I can't see an `eo:epsg` in [the stac repo](https://github.com/radiantearth/stac-spec/search?q=eo%3Aepsg&unscoped_q=eo%3Aepsg) -- should it be `proj:epsg` or something? 

(and presumably this will be None on datasets with a non-epsg crs?)

cubedash/templates/dataset.html
```diff
@@ -161,11 +161,11 @@ <h2>Raw Metadata</h2>
...
                 center: [-26.2756326, 134.9387844],
                 layers: [osm_carto_light, dataset_data],
-                zoomControl: false,
-                attributionControl: false,
+                zoomControl: true,
+                attributionControl: true,
                 scrollWheelZoom: false
             });
-            L.control.zoom({position: "bottomright"}).addTo(map);
+            // L.control.zoom({position: "bottomright"}).addTo(map);
```
jeremyh: The attribution is currently being put in the page footer instead of each individual map, so that it's less noisy, see here: https://github.com/opendatacube/datacube-explorer/blob/0eced1737952ecbdc18294a628a782cde3508741/cubedash/templates/layout/base.html#L63
Do you think we need to change it?

The zoom control is being added manually so that it's right-aligned on the map instead of left-aligned. Because otherwise the alignment looks off against the rest of the page:

image

(I should have added comments to make these look less like accidents)

cubedash/_stac.py
alexgleith: I was copying it from the Landsat STAC implementation, see here: https://landsat-stac.s3.amazonaws.com/landsat-8-l1/147/032/2015-01-02/LC81470322015002LGN00.json

But I note that it isn't in the latest `eo` extension, here: https://github.com/radiantearth/stac-spec/tree/master/extensions/eo. So it probably should go.

alexgleith: Ah, looks like it was there in the 0.6.0 version of the `eo` extension: https://github.com/radiantearth/stac-spec/tree/v0.6.0/extensions/eo

requirements-test.txt
```diff
@@ -32,7 +32,7 @@ colorama==0.4.1
 constantly==15.1.0
 coverage==4.5.4
 cssselect==1.1.0
-dask==2.3.0
+dask
```
whatnick: @alexgleith For evironment setup not a fan of unpinned versions...

requirements-test.txt
```diff
@@ -65,7 +65,7 @@ mccabe==0.6.1
 mock==3.0.5
 more-itertools==7.2.0
 munch==2.3.2
-netCDF4==1.5.1.2
+netCDF4
```
whatnick: Same here.. what is a working version

apt-requirements.txt
```diff
@@ -0,0 +1,18 @@
...
+libgdal-dev
+libhdf5-serial-dev
+liblapack-dev
+libnetcdf-dev
+libproj-dev
+libudunits2-0
+libudunits2-dev
+libxml2-dev
+libxslt-dev
+postgresql-10-postgis-2.4
```
whatnick: Use 11 for server compatibility

.travis.yml
```diff
@@ -39,30 +36,29 @@ addons:
     - libudunits2-dev
     - libxml2-dev
     - libxslt-dev
-    - postgresql-9.6-postgis-2.4
+    - postgresql-10-postgis-2.4
```
whatnick: Use 11 for server compatibility

.travis.yml
```diff
@@ -11,25 +11,22 @@ git:
...
-env:
 matrix:
   include:
   - env: STABLE_DATACUBE=true
   allow_failures:
   - env: STABLE_DATACUBE=''
 
 addons:
-  postgresql: '9.6'
+  postgresql: '10'
```
whatnick: Use 11 for production server compatibility

whatnick (changes requested):
Let us also pin Werkzeug version to make tests happy

requirements-test.txt
jeremyh: Yeah, I should have caught that. I was keeping all of these pinned for Travis stability, with pyup configured [to bump them periodically](https://github.com/jeremyh/datacube-explorer/pull/35). But I stopped looking at pyup's PRs, so they've falled behind lately.

cubedash/_stac.py
```diff
@@ -323,7 +323,7 @@ def as_stac_item(dataset: DatasetItem):
             **dict(_build_properties(dataset.odc_dataset.metadata)),
             "odc:product": dataset.product_name,
             "odc:processing_datetime": utc(dataset.creation_time),
-            "cubedash:region_code": dataset.region_code,
+            "cubedash:region_code": dataset.region_code
```
jeremyh: I think `black` is failing on this line

jeremyh (approved):
Nice :+1:

whatnick (approved):
👍

Issue #81, opened 2020-02-26
Fix build/test for werkzeug/floating point noise


whatnick:
Werkzeug version has fallen out of sync. Polygon equality check is too rigid and does not account for floating point noise.

whatnick:
Fixed

PR #82, opened 2020-02-26
#81 Fix werkzeug support and tests

whatnick:


jeremyh (approved):
Thanks Tish! Looks good. :+1: 

I broke our antimeridian handling locally and the test still picked it up, so it seems a-ok to me.

The werkzeug fix is still causing problems for me in some situations (such as when I run one pytest by itself). But I think a more thorough fix is something we can do later.

PR #86, opened 2020-02-27
Use a flask-themes that we control

jeremyh:
Supercedes #85, as discussed with @whatnick  on slack.

We forked flask-themes into a version we control, and patched it so that we no longer need the werkzeug fix. The original is maintained very sporadically, so this seemed worth it.


whatnick (approved):
👍

Issue #87, opened 2020-02-27
Add DE Africa Logo


whatnick:
Current theme uses Australia Branding for Africa.

![image](https://user-images.githubusercontent.com/491396/75411355-15cba480-5973-11ea-839f-b31ecfcb5645.png)

Africa specific assets [here](https://intranet.ga.gov.au/everyday-tasks/communication-and-design/templates/priority-projects/Digital-Earth-Africa_DE-Africa.ZIP)

alexgleith:
This was fixed in #88

PR #88, opened 2020-02-27
Start an Africa theme

jeremyh:
I started looking at implementing colour theming but that wont be finishable today. So here's a start with just the logo.

Can be enabled by setting theme to 'deafrica' in your [settings](https://github.com/opendatacube/datacube-explorer#how-can-i-configure-the-deployment), ie.

    $ cat settings.env.py 
    CUBEDASH_THEME = 'deafrica'
    $

The spacing around the logo isn't quite perfect in all cases (different screen sizes), so I may have some trivial PRs still upcoming.

The sass project now recommends an npm-based implementation, so I've moved the Makefile to use that too.


jeremyh:
#87, though I didn't tag commits

jeremyh:
africa-theme-initial

alexgleith (approved):
Logo changes look fine!

Don't like committing production css, but it's obviously how we do it for now :-)

PR #93, opened 2020-03-02
Fix Black failures in develop

jeremyh:
To get the build green again until the GH Actions work is ready.

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -12,7 +12,7 @@ Automat==0.7.0
 Babel==2.7.0
 beautifulsoup4==4.8.0
 billiard==3.6.1.0
-black==19.3b0
+black==19.10b0
 boltons==19.1.0
 bs4==0.0.1
 cachetools==3.1.1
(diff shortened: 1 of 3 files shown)


requirements-test.txt
```diff
@@ -12,7 +12,7 @@ Automat==0.7.0
 Babel==2.7.0
 beautifulsoup4==4.8.0
 billiard==3.6.1.0
-black==19.3b0
+black==19.10b0
```
alexgleith, line 15: I'm aiming to make this file redundant. It's best off including this in the `test` requirements in the setup.py :-)

jeremyh: Yep, I'd like to retire it too. A bump is needed in the short-term to fix the current issue: pre-commit is installing the latest Black, but this `requirements-test.txt` is pinning an older Black.

setup.py does has Black in the test requirements already, but doesn't specify anything about versions. Do you think we should? I'm usually hesitant to add versions to setup.py unless it's a hard requirement, and it's not technically used by the tests themselves.

alexgleith (approved):

PR #102, opened 2020-03-03
remove a bunch of installed packages

alexgleith:
Tested and still seems to work! SHould reduce the image size a fair bit.


jeremyh (approved):
Thanks Alex.

It looks like an accident that shellcheck was never turned on in Travis -- we do have some shell files in the repo.

whatnick (approved):
👍  less is more

PR #103, opened 2020-03-03
Fix/sentry version

whatnick:
Try to check codecov with PR from main repo instead of fork.


alexgleith (approved):
Approving in anticipation of the 20 minute tests to work!

Issue #112, opened 2020-03-10
Github Actions fail to extract tag for Docker


whatnick:
Preparation steps fail to capture commit/tag version and push to dockerhub.

![image](https://user-images.githubusercontent.com/491396/76273557-40522180-62d2-11ea-9d73-7f30b862e77f.png)

whatnick:
@alexgleith we can close this ?

PR #113, opened 2020-03-10
#112 Try a tag docker will accept as default

whatnick:
Attempt to make release based tagging work. Deep somewhere in Github actions something is wrong.

### DIFF
--- a/integration_tests/test_version.py
+++ b/integration_tests/test_version.py
@@ -1,0 +1,7 @@
+"""Test this cubedash build has a valid version
+"""
+import cubedash
+
+
+def test_check_version():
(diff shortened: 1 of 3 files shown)


integration_tests/test_version.py
```diff
@@ -0,0 +1,7 @@
+"""Test this cubedash build has a valid version
+"""
+import cubedash
+
+
+def test_check_version():
+    assert cubedash.__version__ != "Not-Installed"
```
alexgleith, line 7: This will behave strangely with Docker in the dev environment, I think, because the code that is being run has _not_ been installed. I'm not sure this test is a good idea.

PR #117, opened 2020-03-16
Change Dockerfile to be based on opendatacube/geobase

alexgleith:
- [x] makes the built image smaller
- [x] should mean that geo libraries are fully working together.

### DIFF
--- a/requirements-docker.txt
+++ b/requirements-docker.txt
@@ -1,0 +1,1 @@
+datacube
--- a/requirements-docker.txt
+++ b/requirements-docker.txt
@@ -1,1 +1,3 @@
+--extra-index-url https://packages.dea.ga.gov.au/
+digitalearthau
 datacube
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,6 @@
 FROM opendatacube/geobase:wheels as env_builder
 COPY requirements-docker.txt /
 RUN env-build-tool new /requirements-docker.txt /env
-RUN pip3 install --extra-index-url \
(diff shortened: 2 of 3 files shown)


jeremyh:
Looks good. That's a huge decrease in runtime -- it's much closer to what Travis was. Oddly, the running of the tests is much faster now, not just the machine creation.

jeremyh (approved):

requirements-docker.txt
```diff
@@ -0,0 +1,3 @@
+--extra-index-url https://packages.dea.ga.gov.au/
+digitalearthau
+datacube
```
Kirill888, line 3: @alexgleith for cloud installs `datacube[s3]`  is needed, but you probably want to enable `performance` and `distributed` as well. There are a lot of timestamps in the metadata and python date parsing is way too generic and hence slow, `performance` option installs a much faster date parsing lib and some other bits too. But it's not just about speed, majority of our environments use that lib so fallback code doesn't get tested all that much with real life work loads.

alexgleith: Distributed would be for Dask, right?

Explorer is not going to be using Dask, and doesn't actually do any data reading to s3.

So what about just installing with the `performance` flag? Or do you think we should install with all three more as a 'best practice' thing?

Kirill888: distributed is for distributed dask, dask itself is not optional.
s3 pulls in boto3, which is probably installed anyway, but yes if you do not read anything from s3 then you don't need it.
performance pulls in ciso8601 and bottleneck, so time parsing and "misc other", not sure how much impact bottleneck has or how it works for that matter.

PR #118, opened 2020-03-17
Enable esri crses

alexgleith:
Fixes a single line of code that tests for an `epsg` string to make it include `esri` strings too, which it turns out are valid in Proj 6.


jeremyh (approved):
:+1: 

We could probably expand that check to allow any `authname:code` pattern -- it's mainly there to stop people using WKT or other crs formats which aren't supported as defaults [yet].

PR #119, opened 2020-03-18
'generate' command should default to --no-init-db

jeremyh:
As suggested by @alexgleith on Slack.

The new prod instances are always passing `--no-init-db` so that they can run with tighter user permissions (presumably only `insert`/`update`, not `create`). 

We could have fixed this accidental requirement of `create` permission, but it's probably better to be consistent with datacube-core/ows and separate `init` as an explicit action.

It will print a 'friendly' red error message if you have not initialised:

```
 $ cubedash-gen 
No cubedash schema exists. Please rerun with --init to create one
```

--- 
I've also included a second commit to clean up some inconsistent stderr/stdout usage in the user messages of the command.

### DIFF
--- a/cubedash/generate.py
+++ b/cubedash/generate.py
@@ -1,11 +1,13 @@
 import multiprocessing
 import sys
 from datetime import timedelta
+from functools import partial
 from typing import List, Optional, Sequence, Tuple
 
 import click
 import structlog
-from click import echo, secho, style
+from click import secho as click_secho
+from click import style
 
 from cubedash.logs import init_logging
 from cubedash.summary import SummaryStore, TimePeriodOverview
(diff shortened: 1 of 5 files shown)


cubedash/generate.py
```diff
@@ -142,7 +142,8 @@ def _load_products(index: Index, product_names) -> List[DatasetType]:
 @click.option("--force-concurrently", is_flag=True, default=False)
 @click.option(
     "--init-database/--no-init-database",
```
alexgleith, line 144: The old option is still there. Is it valid to have both `--init-database` and `--init` ?

jeremyh: Yep -- you can include multiple aliases in click options (eg. '-r' and '--recursive').

I added the shorter version as it seems more suitable and memorable for a frequently used option.

alexgleith (approved):

PR #121, opened 2020-04-27
Remove DEA-specific deployment code and outdated docs

alexgleith:


jeremyh (commented):
Thanks Alex. We might still need the screenshot (perhaps moved to another folder?) -- it's used in the main readme.

alexgleith:
What do you think about screenshot at root, @jeremyh? We could create a `docs` folder... but I don't think it's necessary yet.

jeremyh (approved):
Sure, that works

PR #122, opened 2020-04-27
Implement default extent in options

alexgleith:
In order for the Dashboard not to default to Australian extents, this PR adds a small item to the config so that extents are configurable in the theme.


jeremyh (approved):
Nice, thanks Alex! :slightly_smiling_face: 

Note that the js file is generated from typescript, so I've pushed those changes up to the ts file.

I recommend running [pre-commit](https://github.com/opendatacube/datacube-explorer#pre-commit-setup) too, so you don't need separate "Fix black" commits in the future. Or amend/squash commits afterwards.


ps. I didn't copy the console.log messages as it's slightly misleading ([logging almost always](https://github.com/opendatacube/datacube-explorer/blob/d0a1c3d787cfffe0fa4bc6fdf36a5a3d2e9a1495/cubedash/templates/overview.html#L7)) and isn't visible to the user anyway.

jeremyh:
> Thanks. We should get Typescript working for... others!

Yeah, sorry. I'll add it to the makefile. (PyCharm has a "recompile on changes" toggle, which just runs `tsc` on the file directly. No other special setup. It also autodetects and hides the js/map files so it's hard to accidentally change compiled outputs)

PR #125, opened 2020-05-06
Update for stac 0.9 schema changes

jeremyh:
This still does not include some of the optional stac extensions that we want to use (like eo, grid...), but the core of it is now conformant to 0.9.0's schema changes, so I think it's worth pushing anyway.

The tests are expanded a little too, and validate against the updated schemas from the stac repo.


alexgleith:
Hey @jeremyh this is looking good!

I wonder if we should be using this tool to create documents https://github.com/sat-utils/sat-stac

Also, we're really hoping to target 1.0-beta with Proj and EO extensions, as this is what the Sentinel-2 data will use. And the USGS are aiming for this too. It's captured in the [dev](https://github.com/radiantearth/stac-spec/tree/dev) branch here.

jeremyh:
Yep, 1.0 is the goal.

I figured this was enough of a working milestone to merge, as it's an improvement over the current head, and matches the current "release" of stac.

(I also usually prefer smaller PRs and not keeping branches separate for too long -- especially when it's a secondary priority like this and frequently being forgotten.)

jeremyh:
sat-stac looks ok :+1: . It doesn't contain anything for writing yet that we could use, but might be a place we could contribute back to in the future.

alexgleith (approved):

PR #138, opened 2020-06-04
Refactor github actions into stage based smaller steps

### LINKED ISSUE #135: Refactor GHA Pipelines into more granular steps
Continue work on this branch : https://github.com/opendatacube/datacube-explorer/tree/jez/gha 

Use more granular steps :
- Lint and Test on all branches and PR's
- Docker build and Push only on develop

whatnick:
Closes #135


alexgleith:
Separating linting and testing means that you're building the image twice and both times without a cache... that'll make it a lot slower!

whatnick:
> Separating linting and testing means that you're building the image twice and both times without a cache... that'll make it a lot slower!

But they run in parallel so may be fine, and testing takes a lot longer anyway. If @jeremyh would like it this way ( I suspect he wants separate badges) , I am happy to play along.

jeremyh:
> Separating linting and testing means that you're building the image twice and both times without a cache... that'll make it a lot slower!

Yeah. It should be parallel like Tish mentioned. I don't feel strongly about it, but saw some other projects doing it (like Black). They probably have lighter Docker builds though.

I do like the change that lint errors will not block the tests from running. A few times now the two have broken at the same time, and the person fixing doesn't realise, causing an extra iteration of fix-push-wait-for-build.

jeremyh (approved):

PR #143, opened 2020-06-15
A workaround for non-epsg strings using fuzzy match

### LINKED ISSUE #127: Explorer does not deal with non-EPSG style CRS's
Datacube allows older metadata styles where full CRS is present in the metdata instead of the AUTHORITY:NUMERIC-CODE style explorer expects. Use backwards lookup using PyPROJ database for these. Specific case / linked to #30 

weird_crs

whatnick:
Closes #127 for now.


jeremyh (approved):
:+1:

Issue #145, opened 2020-06-23
Explorer new theme feedback for action


pindge:
## Visual Issues
 - dataset overview on the right hand side `eo` should be `EO`
 - ~~fix dataset overview bottom white (100px padding) space~~
 - ~~need to fix `overflow: scroll` where the scroll bar is splitting the page~~
 - ~~reduce footer height, merge all footer text into one line~~
 - think about menu bar height and maybe shorten the width to allow space for map
 - remove `_` from page title and dataset name
 - ~~remove shadow from logo~~
## Link issues - broken links or misleading links
 - is there an actual navigation issue?
## Usability issues
 - ~~left-hand side date picker needs to disable month with no data~~ - > split into https://github.com/opendatacube/datacube-explorer/issues/155
 - unknown whitebox above `recenter` on the map
 - Add platform `NCI/AWS/Sandbox` information to title
 - product selector too noisy, underline + bold + italics + bold italics  white text on blue
## Feature wishlist - what links/features should be available but it is not
 - Add a `technical` section in the menu area
 - introduce animation to allow space splitting when clicked, address the issue with too much white space
## For further discussion

omad:
Well done, huge improvement!

Not sure how you want to handle feedback, but the first thing for me is:
* lots of unnecessary scroll bars displayed in Firefox.

PR #148, opened 2020-06-25
Theme rework

whatnick:
Major rework on the explorer front end needs to be stabilised to a first acceptable cut then iterated on. Drafting this is as place to collect reviews on this specific change and make future changes incremental. First step towards #145 .

review/demo link: https://explorer-theme-demo.dev.dea.ga.gov.au/aster_false_colour


jeremyh:
Sorry I've been slow to review this -- it looked large. Now that NCI is down for maintenance I should have some time.

jeremyh (commented):
Thanks Pin. It looks like a nice start. I like the new header a lot.

A few things I noticed:


- Narrow-screen display has quirks (the map disappears). The current Explorer (see right, below) switches to a vertically-stacked layout when screen is thin — I suspect some of the same css is still adjusting on thin layout, causing the quirk?





(iPhones, iPads suffer from this; see bottom)


- There’s no scroll-bars on any system. This was a problem once before for Explorer’s menus, but in the refresh it’s affecting everything since they’re all separate frames. Some users are not able to scroll at all without scroll bars.


- All 404 pages are showing “no datasets found” message rather than their own error messages.


- The last item(s) in the sidebar are never visible on various pages, because they're hidden by the footer. 

    - (Adding a slight margin to the bottom is also a nice hint to the user that they’ve reached the end of scrolling. We used double horizontal padding previously.)

- I’d mildly lean against introducing in-line styles in the html elements ( `style="margin-bottom:0px;"`) to try to keep styling less fragmented.

**Top Bar**




- ~~The `/about` page (“Overview”) was hidden previously because it has nothing useful in it. It has a list of products, but those are already in the side menu. I’d rather delete or improve it rather than expose it in its current form as it’s just noise.~~ Done in #192, #199

 

- I like the idea of making the API endpoints more discoverable, like the “View product Iists.txt”, though I don’t think it should be displayed so prominently. In this design it’s given more visual prominence than the main menu options but isn’t really intended as a web-viewable-page. The information it contains is redundant (the menu has a product list) so is only really intended for API usage.

    Perhaps this is the sort of thing that could be listed on the about/overview page, along with the other API endpoints?

- The “home” link on this menu is always the active one, even when viewing other page(s). ("active" is being styled as faded)


- The blue square in the DEA logo only has a slight outline, which looks like a mistake? What do our other apps do here? On large screens it makes it look like Western Australia is missing.






----------

 
### Menu

These in particular are points I probably care about more than other people, because I’m a pedant. But I think consistent visual language makes apps easier to use. If the answer is “leave this for Jeremy to fix later” that’s okay with me 🙂 









- The menu has many different inconsistent paddings and spacings. Some mouse-hover effects have padding and some not. Some underline on hover and some don’t.

- There’s a lot of different horizontal alignments. They’re all nearly-the-same but not quite, which looks like a mistake? The menu, items, sub-items, logo etc. The dataset counts look like they never align (they’re right-aligned but the variable width and wrapping makes them not look so).


    Eg. left side of the logo doesn’t align horizontally on any of the pages any more:



- The group name is underlined but is not a link (clickable). Everything else on pages that are underlined are links.
group-heading
    - (the `/about` page has this the other way around in its list: product links are clickable and underlined, group names are not and not.)    
    
- The menu doesn’t visually hint that it’s expanded, or that it's clickable to collapse. 
    - It feels like it’s replaced when opened rather than expanded, which makes this less clear too. 
    - The year/month menus are not usable/visible when it’s open, and it’s not obvious that you need to close the menu to find them. 
    - The currently active product looks duplicated because it’s both at the top and in the expanded menu (this makes sense for pop-up-sub-menus but not the expanded-style menus that it’s now using).

    One possibility here is to open the sub-menu to the right, rather than over the top of existing entries.

- If we do want to keep the dataset count in this list, maybe "34 datasets" instead of  "dataset count: 34", to match the other counts on the page


- The visual hierarchy<sup>[(1)](https://www.interaction-design.org/literature/topics/visual-hierarchy)</sup> feels off to me, making it harder to scan: the menu group items have more emphasis than the currently-selected-product. And less-important info like dataset count are emphasized the same as product names The product name is the key thing people scan for when choosing a product, so is worth unique distinction.

- The new elements all have a lot more emphasis than the content of the page. We might want to tweak the older, more-desaturated colours of Explorer (eg, the polygons) to match the new highly-saturated colours. Or fiddle with the balance somewhere.



----------


- The upper-case `metadata_type` name is inconsistent with all the lower-case product names elsewhere on the page:


(they’re displayed in that way as that’s how they’re represented in ODC (APIs, dev tools, config etc) — these names are used as pseudo-ids in ODC)


----------


### iPhone view
Narrow screen collapse




----------

### iPad view
Has weird horizontal scrolling?




----------

Nice work overall Pin. I’m excited about sprucing some of this stuff up.  

Feel free to come talk if you want to discuss anything.

whatnick:
@jeremyh there are some notes around the `about` page. Shall we revise these in the light of its elevated status.

jeremyh:
> @jeremyh there are some notes around the `about` page. Shall we revise these in the light of its elevated status.

Yep! I've struck that line out, it's been improved.

pindge:
I am going to close this mega PR, I will raise smaller incremental THEME WORK prs.

PR #151, opened 2020-06-30
Fix issue with incrementing ids

alexgleith:
* Updates the fields that are included in the update part of the upsert comment


cubedash/summary/_stores.py
```diff
@@ -374,6 +374,7 @@ def _set_product_extent(self, product: ProductSummary):
             for name in product.derived_products
         ]
         fields = dict(
+            name=product.name,
```
jeremyh, line 377: I preferred not including this in the update statement, as `name` should never change. Though I'm probably just superstitious.

jeremyh (approved):
Thanks Alex.

cubedash/summary/_stores.py
jeremyh: (I also had vague memories of there being a performance downside. But this query isn't run often enough for that to matter...)

alexgleith: Instead of getting the ID by name and then updating on ID, we could do the update by name and handle the exception and insert then... saves a query. Then we're updating on name only.

Issue #144, opened 2020-06-21
Product id sequence can overflow


whatnick:
Cubedash gen fails in the Geoscience Australia [dev deployment](https://explorer.dev.dea.ga.gov.au/s2a_nrt_granule/2020/5/24) and presumably in production. Failure occurred about a month ago , showing the delayed NRT products in the link above however recent improvements to logging have surfaced them. The NRT streams are functional and OWS can see them :

![image](https://user-images.githubusercontent.com/491396/85236381-e0d2e880-b460-11ea-92d2-5c9fd142c8b8.png)

Failure mode is lot of messages as below per product:
```
NOTICE:  identifier "dashgen-fc-percentile-albers-seasonal agdc-1.8.1.dev22+gc2c25d3d" will be truncated to "dashgen-fc-percentile-albers-seasonal agdc-1.8.1.dev22+gc2c25d3"
2020-06-19T07:07:29.650249Z [exception] generate.product.error         product=ga_s2a_ard_nbar_granule
Traceback (most recent call last):
  File "/env/lib/python3.6/site-packages/sqlalchemy/engine/base.py", line 1284, in _execute_context
    cursor, statement, parameters, context
  File "/env/lib/python3.6/site-packages/sqlalchemy/engine/default.py", line 590, in do_execute
    cursor.execute(statement, parameters)
psycopg2.errors.SequenceGeneratorLimitExceeded: nextval: reached maximum value of sequence "product_id_seq" (32767)
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
  File "/env/lib/python3.6/site-packages/cubedash/generate.py", line 46, in generate_report
    store.refresh_product(product, refresh_older_than=refresh_time)
  File "/env/lib/python3.6/site-packages/cubedash/summary/_stores.py", line 181, in refresh_product
    derived_products=derived_products,
  File "/env/lib/python3.6/site-packages/cubedash/summary/_stores.py", line 388, in _set_product_extent
    .values(name=product.name, **fields)
  File "/env/lib/python3.6/site-packages/sqlalchemy/engine/base.py", line 2244, in execute
    return connection.execute(statement, *multiparams, **params)
  File "/env/lib/python3.6/site-packages/sqlalchemy/engine/base.py", line 1020, in execute
    return meth(self, multiparams, params)
  File "/env/lib/python3.6/site-packages/sqlalchemy/sql/elements.py", line 298, in _execute_on_connection
[...]

PR #152, opened 2020-06-30
#144 Test harness for upsert index increment

whatnick:
Check that last_value from `force-refresh` does not increment. Why it is **74** to start with I am not sure.


alexgleith (approved):

alexgleith (approved):

alexgleith:
Nice one!

PR #154, opened 2020-06-30
#153 Use git to get unstable semver

### LINKED ISSUE #153: Add githash to Semvers for unstable docker images from master
When PR's are merged to master push Docker images with tag following the pattern
```
<tagged-release>-<commit-count>-<githash>
```

This can be generated using:
```
git describe --tags
```

whatnick:
Closes #153 
- Use git to fetch unstable semver
- Push unstable Semver to dockerhub
- Scan the latest unstable semver with Trivy instead of latest
- Scan release semver with Trivy ( more scanning more security ? )

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -34,32 +34,28 @@ jobs:
       - name: Log the unstable tag
         run: echo $UNSTABLE_TAG
 
-      - name: Build and Push latest Docker image
+      - name: Build and Push unstable + latest Docker image tag
+        if: github.event_name != 'release'
         uses: whoan/docker-build-with-cache-action@v4
         with:
           image_name: ${{ env.IMAGE_NAME }}
           username: gadockersvc
+          image_tag: ${{ env.UNSTABLE_TAG }},latest
           password: "${{ secrets.DockerPassword }}"
           build_extra_args: "--build-arg=ENVIRONMENT=deployment"
 
-      - name: Build and Push unstable Docker image tag
-        uses: whoan/docker-build-with-cache-action@v4
-        with:
-          image_name: ${{ env.IMAGE_NAME }}
-          username: gadockersvc
-          image_tag: ${{ env.UNSTABLE_TAG }}
-          password: "${{ secrets.DockerPassword }}"
-          build_extra_args: "--build-arg=ENVIRONMENT=deployment"
-
-      - name: Build and Push latest Docker image (old name)
+      - name: Build and Push unstable + latest Docker image (old name)
+        if: github.event_name != 'release'
         uses: whoan/docker-build-with-cache-action@v4
         with:
           image_name: ${{ env.OLD_IMAGE_NAME }}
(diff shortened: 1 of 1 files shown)


jeremyh (approved):

.github/workflows/docker.yml
```diff
@@ -35,6 +42,15 @@ jobs:
           password: "${{ secrets.DockerPassword }}"
           build_extra_args: "--build-arg=ENVIRONMENT=deployment"
 
+      - name: Build and Push unstable Docker image tag
+        uses: whoan/docker-build-with-cache-action@v4
+        with:
+          image_name: ${{ env.IMAGE_NAME }}
+          username: gadockersvc
+          image_tag: ${{ env.UNSTABLE_TAG }}
```
alexgleith: We want to push `latest` here too

.github/workflows/docker.yml
```diff
@@ -35,6 +42,15 @@ jobs:
           password: "${{ secrets.DockerPassword }}"
           build_extra_args: "--build-arg=ENVIRONMENT=deployment"
 
+      - name: Build and Push unstable Docker image tag
+        uses: whoan/docker-build-with-cache-action@v4
```
alexgleith: I think we want `if != 'release'`

.github/workflows/docker.yml
```diff
@@ -35,6 +42,15 @@ jobs:
...
+        with:
+          image_name: ${{ env.IMAGE_NAME }}
+          username: gadockersvc
+          image_tag: ${{ env.UNSTABLE_TAG }}
+          password: "${{ secrets.DockerPassword }}"
+          build_extra_args: "--build-arg=ENVIRONMENT=deployment"
+
       - name: Build and Push latest Docker image (old name)
```
alexgleith: Let's get rid of the old docker image... moving forwards!

.github/workflows/docker.yml
whatnick: Latest is the default. Pushed in another step. Not sure if the action supports multiple tags. Looking at it in a bit.

whatnick:
```
image_tag: Tag(s) of the image to build. Allows multiple comma-separated tags (e.g. one,another) (default: latest).
```
Squashing into one step.

.github/workflows/docker.yml
whatnick: Too many changes I am not sure I have changed them all in all my environments. Delayed deprecation.

jeremyh:
I shouldn’t review parts of the system I have little experience with 🙂

whatnick:
> I shouldn’t review parts of the system I have little experience with 🙂

Inducting / Subverting you into the ways of the docker. But just keeping you aware of the changes we are making. Need that increment change tested and shipped out soon to keep live dataflows happy. Pushing to dockerhub is sort of the first step to that.

alexgleith (approved):

PR #156, opened 2020-07-08
Improve product date selector header

pindge:
Fix issue https://github.com/opendatacube/datacube-explorer/issues/155

one of the theme rework feedback collected was the usability issue with product date selector.

Instead of making it part of the theme rework, splitting this issue in this PR to make review easier.


jeremyh (approved):
This looks like a nice, simple solution. Thanks Pin. :+1: 

I'd also consider keeping the day/month list full and just greying out the empty ones (like we do with products). But I'm not sure which way is better. This is a big improvement as-is.

PR #159, opened 2020-07-12
Fix chart-timeline-bar misalignment for yearly datasets

pindge:
https://github.com/opendatacube/datacube-explorer/issues/158

I am not sure how to manage the product that is yearly only, maybe we will do a hard pass on `_annaul` and `summary`, then any additional we will just identify in `config` file.

image


pindge:
Closing

PR #160, opened 2020-07-13
Bugfix/none has no items

pindge:
fixes for server error  where `selected_summary.timeline_dataset_counts ` is none.
https://explorer.dev.dea.ga.gov.au/ls5_usgs_l2c1/2002


jeremyh (approved):
👍

Issue #133, opened 2020-06-03
Native eo3 document support


jeremyh:
eo3 documents already work in Explorer thanks to ODC 1.8's automatic conversion (which appends the legacy fields to the document on index), but they don't display very well in Explorer.

This we can improve:

- [x]  the geometry shown in explorer is the square bounding box. We could use the actual footprint available in eo3's native geometry. #162

- [x] support the `region_code` field for displaying grid counts #162
- [x] support for eo3 fields in the document renderer #167
- [x] expand stac api to make use of extra eo3 properties. grids, etc.

jeremyh:
Closing. All mentioned items are done in the current release (2.2.3).

PR #162, opened 2020-07-15
Native support for displaying, summarising EO3

jeremyh:
- Display the dataset geometries for EO3 datasets (instead of just a max/min box).
- Support regions (individual, searchable grid cells with counts and statistics) for all products with a `region_code` (this includes not just EO3, but some older products)
- Update the document renderer for eo3 fields
- Add a sampler that finds product metadata that’s not included in the product definition. These are displayed with a squiggly underline.
- Move some of the product information to a separate product page.

Linked to #133

This includes a DB schema change, so users will need to rerun `cubedash-gen —init` to update their schemas.


jeremyh:
Opening this now for more visibility. It’s been hidden in my own branch for a few weeks.

jeremyh:
This does not include all EO3 features we'd like (especially: stac api improvements). 

But it adds useful functionality, the tests now pass, and the PR is already too large.

(I'll be making a few more smaller PRs soon, so preferably no release immediately)

README.md
```diff
@@ -10,7 +10,7 @@
 
 These directions are for running from a local folder in development. But it will run from any typical Python WSGI server. 
 
-Firstly, install Data Cube. Use of a [Data Cube conda environment](https://datacube-core.readthedocs.io/en/latest/ops/conda.html)
+Firstly, install Open Data Cube. Use of a [Data Cube conda environment](https://datacube-core.readthedocs.io/en/latest/ops/conda.html)
```
alexgleith:
```suggestion
Firstly, install the Open Data Cube. Use of a [Data Cube conda environment](https://datacube-core.readthedocs.io/en/latest/ops/conda.html)
```

cubedash/_utils.py
```diff
@@ -415,6 +437,19 @@ def dataset_shape(ds: Dataset) -> Tuple[Optional[Polygon], bool]:
     return geom, True
 
 
+def to_wrs_84(geometry: BaseGeometry, crs: Union[int, str, dict, pyproj.CRS]):
```
whatnick: Use `to_wgs_84`

cubedash/summary/_schema.py
```diff
@@ -170,6 +176,16 @@
     Column("has_region", Integer),
 )
 
+# The geometry of each unique 'region' for a product.
+REGION = Table(
+    "mv_region",
+    _REF_TABLE_METADATA,
+    Column("dataset_type_ref", SmallInteger, nullable=False),
```
whatnick, line 185: Not to lump in many changes. Shall we bump primary key data type ?

cubedash/_utils.py
alexgleith: Tish pointed this out as it should be `wgs`

cubedash/_utils.py
```diff
@@ -415,6 +437,19 @@ def dataset_shape(ds: Dataset) -> Tuple[Optional[Polygon], bool]:
...
+    origin = pyproj.Proj(init=crs)
+    dest = pyproj.Proj(init="epsg:4326")
+    tranform_wrs84 = functools.partial(pyproj.transform, origin, dest)
```
alexgleith: There are datacube functions that do this, which are faster, somehow!

See: https://github.com/opendatacube/odc-tools/blob/develop/libs/index/odc/index/stac.py#L76

cubedash/_utils.py
```diff
@@ -415,6 +437,19 @@ def dataset_shape(ds: Dataset) -> Tuple[Optional[Polygon], bool]:
...
+    new_geometry = transform(tranform_wrs84, geometry)
+    # Unwrap coordinates, if necessary
+    new_geometry = test_wrap_coordinates(new_geometry)
```
alexgleith: @Kirill888  implemented something for anti-meridean handling too, which is probably better than my effort.

alexgleith (commented):
Impressive piece of work! I haven't gone through it all, but managed to make a couple of suggestions.

whatnick (changes requested):
Small typo and a comment. Looks clean otherwise.

cubedash/summary/_schema.py
jeremyh: The corresponding upstream ODC-core primary key that this links to is a SmallInt, so I'm hesitant to deviate from it. We just need to avoid making bad upsert statements again...

cubedash/_utils.py
jeremyh: Thanks -- that api is much cleaner than doing it ourselves.

Added.

jeremyh:
Moved to ODC's geometry functions -- which involved replacing the test data with more accurate outputs. I checked it by eyeballing the two geometries on a map.

whatnick (approved):

alexgleith (approved):

PR #164, opened 2020-07-26
#96 Tests to cover sentry integration

### LINKED ISSUE #96: Sentry is not usable in Docker deployments
Removing versioneer support broke sentry integration since it has an import of _version.py

whatnick:
Close #96 by covering Sentry code in tests to make sure it is actually called successfully. Work towards upgrading to modern DSN only Sentry SDK.


whatnick:
I have noticed that codecov struggles capturing coverage correctly when 2 files in different folders are named the same in this case `_model.py` . This PR should increase coverage on it but does not.

integration_tests/test_page_loads.py
```diff
@@ -32,6 +33,21 @@ def auto_populate_index(populated_index):
...
+        "dsn": "___DSN___",
+        "include_paths": ["cubedash"],
+    }
+    return client
+
+
+def test_sentry(sentry_client: FlaskClient):
+    rv: Response = sentry_client.get("/", follow_redirects=False)
+    # Redirect to a default.
+    assert rv.location.endswith("/ls7_nbar_scene")
```
jeremyh: (Already discussed on slack, but to record here)

Ideally put a comment to say what we're testing: it currently feels a little strange to have a test body that looks totally unrelated to Sentry, and which could break for unrelated reasons. 

Or even better, if possible, could we ask the sentry lib if it's initialised?

whatnick:
Not having a great time testing this perhaps some [Flask config best practice](https://flask.palletsprojects.com/en/1.1.x/config/#configuration-best-practices)

jeremyh (approved):
:+1:

Issue #169, opened 2020-07-28
Flask themes fallback mode not activating with the latest code


whatnick:
Latest code does not seem to be activating `flask-theme` fallback mode to find templates in the _templates directory. Thus no pages are rendering.

![image](https://user-images.githubusercontent.com/491396/88654203-61e06800-d110-11ea-979f-df3ca1eeba31.png)

jeremyh:
Strange. They're rendering okay locally and in the tests. I've tried several variations of theme settings (including lack of settings) locally and it's working.

Can you post your `settings.env.py`?

I actually don't understand having this error message for that line: nothing is being iterated (except `context`, but it can't be None by Python's rules). Is there a deeper layer in the stacktrace that the message is coming from?

jeremyh:
The line below it is actually more interesting: TemplateNotFound. My suspicion was that it's something environment-related, that the theme folder is not accessible. But I can't reproduce that exact error message just from locally deleting/moving templates or theme folders.

It's also interesting because it's mentioning `_themes/dea/overview.thml` ... but none of Explorer's theme code uses a folder called `_themes`. The string doesn't exist in this repository.

(... and this release didn't change anything with the theme setup, so it's odd that we're seeing this now)

whatnick:
Here is the deployment config hopefully useful:

```
      CUBEDASH_THEME='dea'
      CUBEDASH_DEFAULT_PRODUCTS = ('aster_false_colour','ls5_fc_albers','ls7_fc_albers',)

      # Which field should we use when grouping products in the top menu?
      CUBEDASH_PRODUCT_GROUP_BY_FIELD = 'product_type'

      # Ungrouped products will be grouped together in this size.
      CUBEDASH_PRODUCT_GROUP_SIZE = 5

      # Maximum search results
      CUBEDASH_HARD_SEARCH_LIMIT = 100

      # Maximum number of source/derived datasets to show
      CUBEDASH_PROVENANCE_DISPLAY_LIMIT = 30

      # Customise '/stac' endpoint information
      STAC_ENDPOINT_ID = 'AWS_EKS_Dev_Explorer'
      STAC_ENDPOINT_TITLE = 'AWS eks dev explorer'
      STAC_DEFAULT_PAGE_SIZE = 20
      STAC_PAGE_SIZE_LIMIT = 1000

      CUBEDASH_DEFAULT_SCHEMA = 'cubedash'

      # Sentry access
[...]

PR #167, opened 2020-07-28
Support EO3 raw documents, many improvements to document handling

jeremyh:
- Cleaner EO3 documents

Make Explorer strip compatibility fields from EO3 Datasets before display+download, so they're identical to the originals.

- New API Endpoints for downloading raw yaml documents:

```
/dataset/<dataset-id>.odc-metadata.yaml
/product/<name>.odc-product.yaml
/metadata-type/<name>.odc-type.yaml
```

These are convenient to use with ODC tools:

    datacube dataset add 'http://localhost:8080/dataset/1264509d-f5ad-5f6d-8bc3-f2c7c41f59ba.odc-metadata.yaml'

- Fix the display of nested dicts/arrays in documents
- Make large arrays collapsable, and collapsed by default:

Screenshot from 2020-07-28 15-25-24

- Add a "View RAW" button next to all documents (products, metadata-types, datasets), to download the plain-text yaml document:

Screenshot from 2020-07-28 15-27-05

- Include `location` in downloaded EO3 dataset documents, so they can match the original index.

PR #170, opened 2020-07-29
Add tests for legacy-db support & a page-test cli-tool

jeremyh:
These reproduce #169 on my system.

... and they're fixed [accidentally] by the code clean-up in #167 

This also adds functions that will allow us to experiment more with tests of Explorer in unusual
Index situations. You can modify the Index, and then "load all pages to see if any fall over".

There's also a second commit which fixes another edge-case it found.


jeremyh:
Extracted the test-all-pages function into a cli tool called `cubedash-page-test`: https://asciinema.org/a/7WRNe4efZFUyfchWRBUXwb3hJ

This is useful for running against local copies of "real" datacubes, like Africa and NCI, which are more useful tests than our test data alone. In theory, you could run it against a remote Explorer too...

jeremyh:
Converting back to a draft -- that's a _big_ increase in test run time. (it isn't as big locally). 

I might need to combine some of the test data loads.

jeremyh:
A new release of Black came out, and we're unfortunately not pinning the version of Black in our Docker container so it's broken the lint build. Added a commit to move to the new version.

jeremyh:
The tests are now (slightly) faster than without this PR, despite adding meaningful (non-%) coverage. Marked as ready for review. :tada:

alexgleith (approved):

PR #174, opened 2020-08-03
Prometheus Metrics for deployment benchmarking

### LINKED ISSUE #173: Enable request monitoring in production deployments with Prometheus
Use [Flask Prometheus](https://github.com/rycus86/prometheus_flask_exporter) to capture response time and error codes in staging / production deployements in Kubernetes.

whatnick:
Closes #173 . Note usage of lower case prometheus os environment variables.

### DIFF
--- a/cubedash/_model.py
+++ b/cubedash/_model.py
@@ -8,7 +8,6 @@ import flask
 import flask_themes
 import structlog
 from flask_caching import Cache
-from prometheus_flask_exporter.multiprocess import GunicornInternalPrometheusMetrics
 from shapely.geometry import MultiPolygon
 
 # Fix up URL Scheme handling using this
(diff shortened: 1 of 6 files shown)


cubedash/_model.py
```diff
@@ -7,6 +8,7 @@
 import flask_themes
 import structlog
 from flask_caching import Cache
+from prometheus_flask_exporter.multiprocess import GunicornInternalPrometheusMetrics
```
jeremyh: It's listed as an optional dependency, so the import should be optional too. Perhaps do the same thing as the raven import.

setup.py
```diff
@@ -13,6 +13,7 @@
     "pytest",
     "pytest-benchmark",
     "pytest-cov",
+    "prometheus_flask_exporter",
```
jeremyh: This one has underscores

setup.py
```diff
@@ -33,6 +34,7 @@
         # Monitoring
         "raven",
         "blinker",
+        "prometheus-flask-exporter",
```
jeremyh, line 37: This one has dashes.

(one of them is wrong?)

jeremyh (approved):

PR #176, opened 2020-08-05
Add an option to recreate existing dataset extents

jeremyh:
...and do it automatically as part of the schema update, so that the newer, better extents will show up.

Option:
```bash
$ cubedash-gen --recreate-dataset-extents <products...>
```

I've also changed `generate` so that it doesn't spawn any subprocesses at all when workers==1. This makes debugging a little nicer.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,12 +6,28 @@ help: ## Display this help text
 .PHONY: install
 install: ## Install all requirements and explorer
 	pip install -U setuptools pip
-	pip install -U -r requirements-test.txt
 	pip install -e .[test]
+
+.PHONY: install-flake8
+install-flake8:
+	pipx install flake8
+	# Add the same plugins as the .pre-commit-config.yaml
+	pipx inject flake8 \
+		   dlint \
+		   flake8-broken-line \
+		   flake8-bugbear \
+		   flake8-builtins \
+		   flake8-coding \
+		   flake8-debugger \
+		   flake8-executable \
+		   flake8-logging-format \
+		   flake8-pep3101 \
+		   flake8-pytest-style \
+		   flake8-pytest \
+		   flake8-rst-docstrings
 
 .PHONY: format
 format: ## Reformat all Python code
-	isort -rc cubedash integration_tests
 	black cubedash integration_tests ./*.py
 
 .PHONY: lint
--- a/README.md
+++ b/README.md
@@ -33,6 +33,8 @@ Initialise and create product summaries:
 
 (This can take a long time the first time, depending on your datacube size.)
 
+Other available options can be seen by running `cubedash-gen --help`.
+
 ### Run
 
 Explorer can be run using any typical python wsgi server, for example:
--- a/cubedash/generate.py
+++ b/cubedash/generate.py
@@ -51,7 +51,7 @@ def generate_report(
         log.info("generate.product.refresh.done")
 
         log.info("generate.product")
-        updated = store.get_or_update(product.name, None, None, None, force_refresh)
+        updated = store.get_or_update(product.name, force_refresh=force_refresh)
         log.info("generate.product.done")
 
         return product_name, updated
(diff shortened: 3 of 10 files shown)


whatnick (commented):
Any README.md or docs entries for the fancy all singing all dancing cubedash-gen ?

jeremyh:
I still have 'add readthedocs site' on my list for before the release, for higher-level overview.

But I've expanded a few of the explanations in the `--help` text:
```
$ cubedash-gen --help
Usage: cubedash-gen [OPTIONS] [PRODUCT_NAMES]...

  A tool to generate and update Explorer's caches of datacube data.

  Explorer's view of ODC data is too expensive to calculate on every page
  load (or API call), so Explorer uses its own tables to calculate this
  summary information ahead-of-time.

  The cubedash-gen command creates these summaries in a schema called
  `cubedash`, separate from datacube’s own schema.

  By default, only missing summaries will be added for the specified product
  names; and it will not recreate summaries that already exist.

  ---

  Datacube config

  To choose which datacube to point to, it takes identical datacube config
  (-C) and environment (-E) options as the `datacube` command, and reads
  identical datacube config files and environment variables.

  ie. It will use the datacube that is shown by running the command
  `datacube system check`

  See datacube’s own docs for this configuration handling.

  ---

  Examples

  Create Explorer's schemas and generate all summaries that don't exist:

      cubedash-gen --init --all

  Recreate all summaries for two products:

      cubedash-gen --force-refresh ls8_nbart_scene ls8_level1_scene

  Drop all of Explorer’s additions to the database:

      cubedash-gen --drop

Options:
  -E, --env TEXT
  -C, --config, --config_file TEXT
  --all                           Refresh all products in the datacube, rather
                                  than the specified list.

  -v, --verbose                   Enable all log messages, instead of just
                                  errors.
                                  
                                  Logging goes to stdout unless `--event-log-
                                  file` is specified.
                                  
                                  Logging is coloured plain-text if going to a
                                  tty, and jsonl format otherwise.

  -j, --jobs INTEGER              Number of concurrent worker subprocesses to
                                  use (default: 3)
                                  
                                  This should match how many io-and-cpu-heavy
                                  queries your DB would like to handle
                                  concurrently.

  -l, --event-log-file PATH       Output log messages to file, in jsonl format
  --refresh-stats / --no-refresh-stats
                                  Refresh general statistics tables that cover
                                  all products (default: true)
                                  
                                  This can be slow, and only needs to be done
                                  once (at the end) if calling cubedash-gen
                                  repeatedly

  --force-refresh / --no-force-refresh
                                  Force all time periods to be regenerated,
                                  rather than just the missing ones.
                                  
                                  (default: false)

  --recreate-dataset-extents / --append-dataset-extents
                                  Rebuild Explorer's existing dataset extents
                                  rather than appending new datasets.
                                  (default: false)
                                  
                                  This is useful if you've patched datasets or
                                  products in-place with new geometry or
                                  regions.

  --force-concurrently            Refresh materialised views concurrently in
                                  Postgres. (default: false)
                                  
                                  This will avoid taking any locks when
                                  updating statistics, but has some caveats,
                                  see https://www.postgresql.org/docs/10/sql-
                                  refreshmaterializedview.html

  --init-database, --init / --no-init-database
                                  Create Explorer's schemas, and prepare the
                                  database, before doing anything.

  --drop-database, --drop         Drop all of Explorer's database additions
                                  and exit.

  --help                          Show this message and exit.
```
(the `-E/--env` and `-C/--config` options come from datacube-core, so I can't add help text to them directly here.)

jeremyh:
I should also mention I added the `--drop` command too:  "Drop all of Explorer's database additions and exit." (mirroring the `--init` command)

It's used internally by the tests, and seemed worth exposing publicly on the cli as well.

jeremyh:
Squashed a few commits for cleaner history.

I've added a mention of `cubedash-gen --help` in the readme, hopefully that's good enough for now.

whatnick (approved):
💯 let's build this and iterate with deployments in staging.

Issue #172, opened 2020-07-31
Post-union geometry errors should be handled more cleanly


jeremyh:
Error posted by @whatnick on Slack, from an Africa generate failure:
```
Traceback (most recent call last):
  File "/env/lib/python3.6/site-packages/sqlalchemy/engine/base.py", line 1284, in _execute_context
    cursor, statement, parameters, context
  File "/env/lib/python3.6/site-packages/sqlalchemy/engine/default.py", line 590, in do_execute
    cursor.execute(statement, parameters)
psycopg2.errors.InternalError_: GEOSUnaryUnion: TopologyException: Input geom 0 is invalid: Self-intersection at or near point -13943223.526815126 -4216248.3542871606 at -13943223.526815126 -4216248.3542871606


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/env/lib/python3.6/site-packages/cubedash/generate.py", line 50, in generate_report
    updated = store.get_or_update(product.name, None, None, None, force_refresh)
  File "/env/lib/python3.6/site-packages/cubedash/summary/_stores.py", line 569, in get_or_update
    force_refresh=True,
  File "/env/lib/python3.6/site-packages/cubedash/summary/_stores.py", line 611, in update
    for year_ in years
  File "/env/lib/python3.6/site-packages/cubedash/summary/_model.py", line 62, in add_periods
    periods = [p for p in periods if p is not None and p.dataset_count > 0]
  File "/env/lib/python3.6/site-packages/cubedash/summary/_model.py", line 62, in <listcomp>
    periods = [p for p in periods if p is not None and p.dataset_count > 0]
  File "/env/lib/python3.6/site-packages/cubedash/summary/_stores.py", line 611, in <genexpr>
    for year_ in years
  File "/env/lib/python3.6/site-packages/cubedash/summary/_stores.py", line 569, in get_or_update
    force_refresh=True,
  File "/env/lib/python3.6/site-packages/cubedash/summary/_stores.py", line 602, in update
    for month_ in range(1, 13)
  File "/env/lib/python3.6/site-packages/cubedash/summary/_model.py", line 62, in add_periods
    periods = [p for p in periods if p is not None and p.dataset_count > 0]
  File "/env/lib/python3.6/site-packages/cubedash/summary/_model.py", line 62, in <listcomp>
    periods = [p for p in periods if p is not None and p.dataset_count > 0]
[...]

Issue #146, opened 2020-06-23
Cubedash-gen failures - self-intersection error for S2_L2A product


alexgleith:
Error is:
>Generating product summaries...
[...]

PR #180, opened 2020-08-21
Convert the Regions materialised-view to a table

jeremyh:
It gives us a lot more flexibility. Such as handling errors and incremental updating.

Users who were running previous unstable/dev versions of Explorer will need to `--init` again.

(we should probably merge #179 before this, so users are properly notified about the needed schema update.)

Planned before merging:

- [x] Improve the handling of broken geometries (the MV suffered from this too)
- [x] Run another full init/upgrade test of Africa DB with this new code on my laptop.

### DIFF
--- a/cubedash/summary/_schema.py
+++ b/cubedash/summary/_schema.py
@@ -1,7 +1,9 @@
 from __future__ import absolute_import
 
-from typing import Set
+import warnings
 from enum import Enum
+from typing import Set
+
 import structlog
 from geoalchemy2 import Geometry
 from sqlalchemy import (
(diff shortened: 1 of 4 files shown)


jeremyh:
Added some broken-geometry handling for #172 #146 . Strangely, ST_MakeValid caused weirdness in some S2 Africa regions:

Screenshot from 2020-09-09 13-40-14

... but Buffering seems to work nicely.

n3h3m (approved):
Looks good to me.

n3h3m (approved):

whatnick (approved):

PR #182, opened 2020-09-04
Regex groups

benjaminleighton:
This PR proposes a regex based grouping mechanism that works on product name.

Current grouping, for example, CUBEDASH_PRODUCT_GROUP_BY_FIELD = 'product_type', defines the grouping of products in the main product selection dropdown in the top menu based on product_type. Product_type is defined in product yaml docs, nominally as an optional 'matching' metadata field. ODC products seem to have little consistency in product_type across products (including none), which leads to a variety of product groupings in Explorer. This PR aims to provide more flexible grouping mechanisms. 

An alternative key() function is implemented that uses a new configuration variable CUBEDASH_PRODUCT_GROUP_BY_REGEX to pass a list of (regex, group name) tuples, which the new key() function operates on if the new env var is set. Co-authored with @mpaget


benjaminleighton:
Oh linting failed. I can try to fix this.

cubedash/_pages.py
```diff
@@ -273,10 +274,34 @@ def _get_grouped_products() -> List[Tuple[str, List[ProductWithSummary]]]:
...
+    if group_by_regex:
+        try:
+            regex_group = {}
+            for regex, group in group_by_regex:
+                regex_group[re.compile(regex.strip())] = group.strip()
+        except Exception:
+            _LOG.warn(
+                "invalid CUBEDASH_PRODUCT_GROUP_BY_REGEX: {}".format(group_by_regex)
+            )
+            group_by_regex = None
```
jeremyh: I think we should just throw a hard error. If they specify a config option it should be valid.

Something like:
```suggestion
        except re.error as e:
            raise RuntimeError(
                f"Invalid regexp in CUBEDASH_PRODUCT_GROUP_BY_REGEX {group!r}: {e!r}"
            )
```
(also catching the specific exception, and giving user the location and message)

README.md
```diff
@@ -128,6 +128,12 @@ You can alter default [Flask](http://flask.pocoo.org/docs/1.0/config/) or
...
     CUBEDASH_DEFAULT_PRODUCTS = ('ls8_nbar_albers', 'ls7_nbar_albers')
     
+    # CUBEDASH_PRODUCT_GROUP_BY_REGEX is a list of tuples of (regex_pattern_string,group_name)
+    # If re.search(regex, product name) then return the corresponding group.
+    # If CUBEDASH_PRODUCT_GROUP_BY_REGEX string parses then regex matching will be used in place of CUBEDASH_PRODUCT_GROUP_BY_FIELD.
+    # If CUBEDASH_PRODUCT_GROUP_BY_REGEX string is malformed an error will be logged and we default back to CUBEDASH_PRODUCT_GROUP_BY_FIELD.
+    # eg "[('.*albers.*','Albers projection'), ('.*level1.*','Level 1 products'), ('.*usgs.*','Level 1 products')]" 
+    CUBEDASH_PRODUCT_GROUP_BY_REGEX = None 
     # Which field should we use when grouping products in the top menu?
     CUBEDASH_PRODUCT_GROUP_BY_FIELD = 'product_type'
```
jeremyh: More consistent language with the other options on the page:

```suggestion
    # Specify product grouping in the top menu.
    # Expects a series of `(regex, group_label)` pairs. Each product will be grouped into the first regexp that matches
    # anywhere in its name. Unmatched products have their own group.
    # eg "(('^usgs_','USGS products'), ('_albers$','C2 Albers products'), ('level1','Level 1 products'), )" 
    CUBEDASH_PRODUCT_GROUP_BY_REGEX = None
    # Otherwise, group by a single metadata field in the products:
    CUBEDASH_PRODUCT_GROUP_BY_FIELD = 'product_type'
```

(Also removed the error handling explanation as suggested below)

cubedash/_pages.py
```diff
@@ -273,10 +274,34 @@ def _get_grouped_products() -> List[Tuple[str, List[ProductWithSummary]]]:
...
+    group_by_regex = app.config.get("CUBEDASH_PRODUCT_GROUP_BY_REGEX", None)
 
-    # Group using the configured key, or fall back to the product name.
-    def key(t):
-        return t[0].fields.get(group_by_field) or t[0].name
+    if group_by_regex:
+        try:
+            regex_group = {}
+            for regex, group in group_by_regex:
+                regex_group[re.compile(regex.strip())] = group.strip()
```
jeremyh: It wont really hurt in this case (as product names can't have whitespace), but I'd rather not strip regexps when they're specified as a precise string literals. If we do want this sort of behaviour I'd rather go all the way and use the [verbose](https://docs.python.org/3/howto/regex.html#compilation-flags) flag.... but I'm not sure.

jeremyh (changes requested):
This is a really nice solution -- thanks guys!

The performance seems fine to me. CPython caches the compiled regexps internally so it's not actually compiling repeatedly.

jeremyh:
Thinking out loud, but potentially the config option could be a dict rather than a list of pairs? They’re ordered now, and cleaner/easier to write.

mpaget:
A list gives an order for selecting which regex matches first. A dict would be fine but regex/groups would have to be unique-like. A use-case for ordered regex/groups might be:

`[("s2_albers", "S2 corrected"), ("s2_nbar", "S2 corrected"), ("s2", "Other S2 products")]`

benjaminleighton:
Thanks @jeremyh I've implemented the suggestions you've made. I didn't know about the cahcing of compiled regexps thanks. Please let me know if there are further changes you'd like.

jeremyh:
> A list gives an order for selecting which regex matches first. A dict would be fine but regex/groups would have to be unique-like. A use-case for ordered regex/groups might be:
> 
> `[("s2_albers", "S2 corrected"), ("s2_nbar", "S2 corrected"), ("s2", "Other S2 products")]`

Yeah, the ordering is definitely useful. We require Py3.6+ which maintains dict order (and is guaranteed to remain the case on all future Pythons).

jeremyh (approved):
:+1: looks good to me. Thanks @benjaminleighton and @mpaget

jeremyh:
Tested locally like this. The examples look nicer when multi-line:

```python
CUBEDASH_PRODUCT_GROUP_BY_REGEX = (
        (r'usgs', 'USGS Products'),
        (r'ga_', 'GA Products'),
)
```

PR #192, opened 2020-09-22
Add a page with behaviour like the cloud-team's product audit table

pindge:
The cloud team is looking to build a dataset audit page to ensure dataset integrity across clusters. Building it on existing explorer to leverage existing functions.

- [x] Add audit page

image

### DIFF
--- a/cubedash/_pages.py
+++ b/cubedash/_pages.py
@@ -266,7 +266,7 @@ def dashboard_page():
             uri = uri_list
         item = {
             "product_name": product.name,
-            "dataset_count": _.dataset_count,
+            "dataset_count": summary.dataset_count,
             "metadata_type": product.definition["metadata_type"],
             "product_metadata": product.definition["metadata"],
             "uri": uri,
--- a/cubedash/_pages.py
+++ b/cubedash/_pages.py
@@ -255,7 +255,7 @@ def dashboard_page():
         uri_list = [
             uri
             for [uri] in dc.index.datasets.search_returning(
-                ["uri"], product=p.name, limit=100
+                ["uri"], product=product.name, limit=100
             )
         ]
 
@@ -265,10 +265,10 @@ def dashboard_page():
         else:
             uri = uri_list
         item = {
-            "product_name": p.name,
+            "product_name": product.name,
             "dataset_count": _.dataset_count,
-            "metadata_type": p.definition["metadata_type"],
-            "product_metadata": p.definition["metadata"],
+            "metadata_type": product.definition["metadata_type"],
+            "product_metadata": product.definition["metadata"],
             "uri": uri,
         }
         dashboard.append(item)
--- a/cubedash/_pages.py
+++ b/cubedash/_pages.py
@@ -258,11 +258,18 @@ def dashboard_page():
         for dl in dataset_list:
             for uri in dl.uris:
                 uri_list.append(uri)
+
+        common_uri = os.path.commonprefix(uri_list)
+        if common_uri:
+            uri = common_uri
+        else:
+            uri = uri_list
         item = {
             "product_name": p.name,
-            "description": p.definition["description"],
             "dataset_count": _.dataset_count,
-            "uri": os.path.commonprefix(uri_list),
+            "metadata_type": p.definition["metadata_type"],
+            "product_metadata": p.definition["metadata"],
+            "uri": uri,
         }
         dashboard.append(item)
 
--- a/cubedash/_pages.py
+++ b/cubedash/_pages.py
@@ -246,7 +246,7 @@ def about_page():
 
 @app.route("/dashboard")
 def dashboard_page():
-    dc = datacube.Datacube()
+    dc = datacube.Datacube(index=_model.STORE.index)
     dashboard = []
     import os
 
--- a/cubedash/_pages.py
+++ b/cubedash/_pages.py
@@ -250,7 +250,7 @@ def dashboard_page():
     dashboard = []
     import os
 
-    for p, _ in _model.get_products_with_summaries():
+    for product, summary in _model.get_products_with_summaries():
         dataset_list = dc.find_datasets(
             product=p.name, limit=100
         )  # sample max 100 datasets
--- a/cubedash/_pages.py
+++ b/cubedash/_pages.py
@@ -251,13 +251,18 @@ def dashboard_page():
     import os
 
     for p, _ in _model.get_products_with_summaries():
-        dataset_list = dc.find_datasets(product=p.name, limit=1)
-        uri = dataset_list[0].uris
+        dataset_list = dc.find_datasets(
+            product=p.name, limit=100
+        )  # sample max 100 datasets
+        uri_list = []
+        for dl in dataset_list:
+            for uri in dl.uris:
+                uri_list.append(uri)
         item = {
             "product_name": p.name,
             "description": p.definition["description"],
             "dataset_count": _.dataset_count,
-            "uri": os.path.commonprefix(uri),
+            "uri": os.path.commonprefix(uri_list),
         }
         dashboard.append(item)
 
--- a/cubedash/_pages.py
+++ b/cubedash/_pages.py
@@ -248,14 +248,16 @@ def about_page():
 def dashboard_page():
     dc = datacube.Datacube()
     dashboard = []
+    import os
+
     for p, _ in _model.get_products_with_summaries():
         dataset_list = dc.find_datasets(product=p.name, limit=1)
-        uri = dataset_list[0].uris[0]
+        uri = dataset_list[0].uris
         item = {
             "product_name": p.name,
             "description": p.definition["description"],
             "dataset_count": _.dataset_count,
-            "uri": uri,
+            "uri": os.path.commonprefix(uri),
         }
         dashboard.append(item)
 
--- a/cubedash/_pages.py
+++ b/cubedash/_pages.py
@@ -244,6 +244,24 @@ def about_page():
     return utils.render("about.html")
 
 
+@app.route("/dashboard")
+def dashboard_page():
+    dc = datacube.Datacube()
+    dashboard = []
+    for p, _ in _model.get_products_with_summaries():
+        dataset_list = dc.find_datasets(product=p.name, limit=1)
+        uri = dataset_list[0].uris[0]
+        item = {
+            "product_name": p.name,
+            "description": p.definition["description"],
+            "dataset_count": _.dataset_count,
+            "uri": uri,
+        }
+        dashboard.append(item)
+
+    return utils.render("dashboard.html", dashboard=dashboard)
+
+
 @app.context_processor
 def inject_globals():
     last_updated = _model.get_last_updated()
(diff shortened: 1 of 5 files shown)


jeremyh:
So the goal is to have an easy view of what products we have, and an example dataset from each?

We might be able to put this on / merge-with the mostly-defunct `/about` page

cubedash/_pages.py
```diff
@@ -244,6 +244,24 @@ def about_page():
     return utils.render("about.html")
 
 
+@app.route("/dashboard")
+def dashboard_page():
+    dc = datacube.Datacube()
```
jeremyh: This will create new connection pools on every page load, and ignore some user config. It's better to reuse the existing ODC index:
```suggestion
    dc = datacube.Datacube(index=_model.STORE.index)
```

cubedash/_pages.py
```diff
@@ -244,6 +244,24 @@ def about_page():
...
+    dashboard = []
+    for p, _ in _model.get_products_with_summaries():
+        dataset_list = dc.find_datasets(product=p.name, limit=1)
+        uri = dataset_list[0].uris[0]
```
jeremyh: Can be empty, which will throw exception

cubedash/_pages.py
pindge: Thanks

cubedash/_pages.py
pindge: thats been changed.

jeremyh:
It's also more a list of products than datasets --  could this table go on the `product-audit` page?

pindge:
> It's also more a list of products than datasets -- could this table go on the `product-audit` page?

I did look at product audit page and it was looking very cluttered, and honestly have no idea what the crosses even represent, so I looked at about page and it had more of that I was looking for, but it uses template global and not that easy to modify. I am not sure where it is best suited.

pindge:
Added way more columns
image

whatnick:
Let us use `dirname` after the common-prefix call and drop the dangling `/S` etc I see in the screenshots.

jeremyh:
I'd probably remove the product-type column as it's not terribly useful these days. Unless the goal is to clean them up in our product definitions :slightly_smiling_face: (something I'd approve of!). `eo3` products would need a fallback to the product-family field which has replaced product-type.

Note that in the newer Explorer release we can group by arbitrary regexps, so we may not need to use product-type for grouping.

cubedash/_pages.py
```diff
@@ -244,6 +244,38 @@ def about_page():
     return utils.render("about.html")
 
 
+@app.route("/dashboard")
+def dashboard_page():
+    dc = datacube.Datacube(index=_model.STORE.index)
+    dashboard = []
+    import os
+
+    for p, _ in _model.get_products_with_summaries():
```
jeremyh: I'm surprised [flake8](https://flake8.pycqa.org/en/latest/) didn't complain about this, but underscore is used for unused variables, and we're using that variable below.

Maybe:
```suggestion
    for product, summary in _model.get_products_with_summaries():
```

cubedash/_pages.py
```diff
@@ -244,6 +244,38 @@ def about_page():
...
+        dataset_list = dc.find_datasets(
+            product=p.name, limit=100
+        )  # sample max 100 datasets
+        uri_list = []
+        for dl in dataset_list:
+            for uri in dl.uris:
+                uri_list.append(uri)
```
jeremyh: This will make page loads quite slow -- we're fetching ~60*100 full datasets from the db.

Since we only care about `uris`, we could fetch just those from the search, rather than the whole datasets:

```suggestion
        # Sample 100 dataset uris
        uri_list = [
            uri
            for [uri] in dc.index.datasets.search_returning(
                ["uri"], product=p.name, limit=100
            )
        ]
```

cubedash/_pages.py
```diff
@@ -244,6 +244,38 @@ def about_page():
...
+
+        common_uri = os.path.commonprefix(uri_list)
+        if common_uri:
+            uri = os.path.dirname(common_uri)
+        else:
+            uri = uri_list
```
jeremyh, line 266: One of these branches sets a string, another a (potentially huge) array. Maybe the latter should just grab one uri as a sample? (but maybe we should tell the user there's no common prefix in these cases)

cubedash/templates/dashboard.html
```diff
@@ -0,0 +1,82 @@
...
+                <tr>
+                    <td>{{ product.product_name }}</td>
+                    <td>{{ product.metadata_type }}</td>
+                    <td><a href="{{url_for('product.raw_product_doc', name=product.product_name, _external=True)}}">{{url_for('product.raw_product_doc', name=product.product_name)}}</a></td>
+                    <td>{{ product.product_metadata.product_type }}</td>
+                    <td>{{ product.dataset_count }}</td>
+                    <td>{{ product.uri }}</td>
+                </tr>
+              {% endfor %}
+            </tbocdy>
```
jeremyh:
```suggestion
            </tbody>
```

cubedash/templates/dashboard.html
```diff
@@ -0,0 +1,82 @@
...
+                    <th>definition doc</th>
+                    <th>Product type</th>
+                    <th>Dataset count</th>
+                    <th>URI</th>
+                </tr>
+            </thead>
+            <tbody>
+                {% for product in dashboard %}
+                <tr>
+                    <td>{{ product.product_name }}</td>
```
jeremyh: Useful as a link
```suggestion
                    <td>{{ product.product_name | product_link }}</td>
```

integration_tests/test_page_loads.py
```diff
@@ -306,6 +306,11 @@ def test_about_page(client: FlaskClient):
     assert b"11 total datasets" in rv.data
 
 
+def test_dashboard_page(client: FlaskClient):
```
jeremyh, line 309: :+1:

cubedash/_pages.py
```diff
@@ -244,6 +244,38 @@ def about_page():
     return utils.render("about.html")
 
 
+@app.route("/dashboard")
```
jeremyh, line 247: The word "dashboard" is used for a lot of things. Maybe we could make it more specific, like `/product-config` or `/product-status` or something?

cubedash/templates/dashboard.html
```diff
@@ -0,0 +1,82 @@
...
+            text-align: left;
+        }
+    </style>
+{% endblock %}
+{% block panel %}
+
+{% endblock %}
+{% block content %}
+    <div class="panel highlight">
+        <h2>Dataset Audit</h2>
```
jeremyh, line 44: Rename header to match the new url... "Product Status" or "Product Overview" or something?

cubedash/templates/dashboard.html
```diff
@@ -0,0 +1,82 @@
...
+        body {
+            width: 100%;
+            max-width: 100%;
+        }
+        thead tr {
+            display: block;
+        }
+        tbody {
+            display: block;
+            height: 700px;
```
jeremyh, line 30: I'm not a huge fan of inner scroll panes when they are the primary content of the page. Many platforms hide scrollbars and normal scroll shortcuts don't work. (see, for example, this github diff page as an example of glorious full-page scrolling :slightly_smiling_face:. Or the [other explorer pages](https://explorer.dea.ga.gov.au/region/ls7_level1_scene/91_80)).

cubedash/templates/dashboard.html
```diff
@@ -0,0 +1,82 @@
...
+            overflow: auto;
+        }
+        table th {
+            width: 220px;
```
jeremyh, line 34: Putting a width on the headers but not body is causing them not to align:
Screenshot from 2020-09-23 11-05-42

I'd rather auto-width as the columns have very different content.

Columns of numbers are also good to right-align for readability.

(If easy, maybe reuse style from the other table-based results pages, eg: https://explorer.dea.ga.gov.au/region/ls7_level1_scene/91_80/2007/8 )

cubedash/templates/dashboard.html
```diff
@@ -0,0 +1,82 @@
...
+    <div class="panel odd">
+        <table>
+            <thead>
+                <tr>
+                    <th>Product Name</th>
+                    <th>metadata type</th>
+                    <th>definition doc</th>
+                    <th>Product type</th>
+                    <th>Dataset count</th>
+                    <th>URI</th>
```
jeremyh: Consistent capitalisation

jeremyh: Perhaps
```suggestion
                    <th>Product</th>
                    <th>Metadata Type</th>
                    <th>Definition</th>
                    <th>Family</th>
                    <th>Dataset Count</th>
                    <th>Location</th>
```

cubedash/_pages.py
jeremyh: or `/dataset-audit` to match header if that's the name we stick with.... but this page is more about products, I think?

cubedash/templates/dashboard.html
jeremyh: Maybe just remove the fixed height, but keep the horizontal scrollable?

pindge: i want the column heading to be always visible, that's my preference, when it comes to the UX part, depends on who actually uses the page and how they feel the information is visible to them, that should be the decision. @whatnick, what do you think.

cubedash/templates/dashboard.html
pindge: this is a flow on affect to enable table header always visible on the page.

jeremyh:
@whatnick  It looks like our comments are blocked on a style disagreement, so we could do with some bike-shedding :slightly_smiling_face: 

Headers that align with the table but scroll away with the page,

Screenshot from 2020-10-06 12-33-11


vs

Headers that don't align, with rows in an inner frame with a separate scroll bar,  so they stay visible as you scroll.

Screenshot from 2020-10-06 12-34-00


I'm biased, and like the readability of the first. But note the rest of Explorer's public pages use the first option so it has consistency going for it. Ideally a solution would be used everywhere. 

Years ago when I had this problem I used javascript to get the best of both worlds, though it looks like the page is now defunct: http://www.transact.com.au/mytransact/tv-guide2

pindge:
> @whatnick It looks like our comments are blocked on a style disagreement, so we could do with some bike-shedding 🙂
> 
> Headers that align with the table but scroll away with the page,
> 
> Screenshot from 2020-10-06 12-33-11
> 
> vs
> 
> Headers that don't align, with rows in an inner frame with a separate scroll bar, so they stay visible as you scroll.
> 
> Screenshot from 2020-10-06 12-34-00
> 
> I'm biased, and like the readability of the first. But note the rest of Explorer's public pages use the first option so it has consistency going for it. Ideally a solution would be used everywhere.
> 
> Years ago when I had this problem I used javascript to get the best of both worlds, though it looks like the page is now defunct: http://www.transact.com.au/mytransact/tv-guide2

in this case, I will remove the sticky header for consistency. the final solution would be to actually build in a table library that all tables can use

cubedash/templates/dashboard.html
whatnick: I like static headers. My use case however does not need us to visually check the page, I will be pulling the contents into a pandas data frame and diffing to ensure the 2 datacubes are in sync. So sort of ambivalent on this particular case.

whatnick (approved):

PR #194, opened 2020-09-23
Assume that regexp-specified product groups are intentional, so don't merge them

benjaminleighton:
Allows single products to be grouped, provides for customization of the default group name.

_merge_singular_groups would merge single groups and partition into CUBEDASH_PRODUCT_GROUP_SIZE. The regex group option may introduce singular groups but these should be presumed intentional and allowed. PR proposes changing _merge_singular_groups to remove merging single groups and keep the partition logic, it also adds a modifiable default group name via settings for non-grouped products.


jeremyh:
Thanks Ben.

This change also affects the default metadata grouping behaviour, as it's no longer combining single groups, and it's repeating the default header name instead of only showing it once:

Previously:
Screenshot from 2020-09-25 10-47-15

Now:
Screenshot from 2020-09-25 10-48-53

With grouping disabled (eg. `CUBEDASH_PRODUCT_GROUP_BY_FIELD = "disabled"`), the header repetition is more apparent too:
Screenshot from 2020-09-25 10-56-31

Previous behaviour:

Screenshot from 2020-09-25 10-58-33


(and it might be surprising to the user that they can't configure this name?)

... I'm undecided on what we want to do here. Maybe this is fine? Maybe I need to work out what would be better default behaviour for people. Most users wont configure regexps, at least initially, so a nice-ish default is preferred.

jeremyh:
I definitely agree with your underlying change though: regexp-specified product groups are intentional from the user, so shouldn't be merged.

benjaminleighton:
Oh gosh these are all very good points, sorry I didn't consider the non-regex CUBEDASH_PRODUCT_GROUP_BY_FIELD behavior enough. I will also think on this. My initial take is that it might be possible to do two things to bring this functionality closer to the previous better single header functionality: 
* Enable combining the _DEFAULT_GROUP_NAME  group so it's header doesn't get repeated. i.e setting the group_name = "" for chunks subsequent to the first
* Use the _DEFAULT_GROUP_NAME instead of "products" even if there are no groups, thus allowing the user to specify it. 
What do you think?

jeremyh:
Those sound okay to me. 

And when _DEFAULT_GROUP_NAME is not specified, I assume it still falls back both to "Products" or "Other Products", depending on whether it's the only group.

jeremyh (approved):
Looks good, thanks @benjaminleighton  :+1: 

It's not combining single-item groups anymore, but I suspect we want to find a better alternative to that behaviour anyway. Or make it opt-in as an option. (we do it because the default 'product_type' field is messily-defined on many products as shown on a screenshot above).

Playing around with it, I think I also want to add the ability to order the groups, as some unimportant groups are sometimes being put first in the list. But that'll be a feature for another PR.  :slightly_smiling_face:

Issue #188, opened 2020-09-17
Add a fallback for eo3 datasets without a geometry field


jeremyh:
I misunderstood the eo3 spec when I implemented #133.

Currently Explorer assumes a dataset will have all three eo3 geo fields (`geometry`, `crs` and `grids`), or none of them (for non-`eo` products like satellite telemetry data). In the latter case Explorer treats the dataset as a non-displayable dataset that has no map view.

But Kirill has mentioned on Slack that the `geometry` field can be missing while `grids` and `crs` are present, and we should then infer the geometry from the union of the coverage of all grids. ODC [implements it this way already](https://github.com/opendatacube/datacube-core/blob/58fabb25683e24d4b3cf738c90049deb689e56a7/datacube/index/eo3.py#L31-L137).

Explorer currently treats any datasets without `geometry` as non-displayable datasets. 
And @alexgleith reported that Africa's clear-pixel counts are affected by this already: https://explorer.digitalearth.africa/ga_s2_clear_pixel_count

We should add a fallback for this case to our Postgis extent extraction.

(eodatasets3 also suffers from this: the `eo3-validate` tool checks for my wrong understanding)

jeremyh:
Fixed in #196, to be included in release 2.2.0

Issue #68, opened 2019-07-31
(finish implementing) incremental-update of summaries


jeremyh:
Currently there's no easy way to update cubedash's product summaries once created.

There is a max_age when the spatial table is updated, but not when generating the summaries themselves, so they are skipped if they exist at all (in the `get_or_update()` function).

We should extend the generate command to allow updating summaries older than a certain age.

(this was never needed previously in the NCI instance as it has a freshly-cloned database every day)

Reported by @dunkgray, as this affected DEA Africa.

alexgleith:
Fixed in: https://github.com/opendatacube/datacube-explorer/pull/84

jeremyh:
Reopening this — we only have a “force” option to force recreation of every summary,  but I’d like to add real incremental updates.

Issue #45, opened 2019-01-08
A creation-time is mandatory for datasets, and this is not made clear to people


jeremyh:
Summary generation throws an exception if datasets are missing a creation time (by default: 'creation_dt' in the yaml)

This has confused a few people, as some datacube guides don't include the field in metadata.

We need to either make it clear why it's failing (a firendly error message?), or else make the field optional in cubedash.

jeremyh:
Quick workaround for Fang was to populate it from the end time ("to time"):

    update agdc.dataset 
        set metadata=jsonb_set(metadata, '{creation_dt}', metadata->'extent'->'to_dt') 
        where metadata->>'creation_dt' is null;

(Note that this was a development machine, you probably don't want to modify your datasets within the index otherwise!)

jeremyh:
Should be fixed in #196. It now falls back to the indexed time of the dataset.

PR #196, opened 2020-09-24
Add EO3 geometry fallback, support dataset archival & make creation-time optional

jeremyh:
- Add a fallback for eo3 datasets without geometry (#188)
- Scan for archived datasets, and remove them from Explorer's extents (one component of #68)
- Make creation-date optional. It will fall back to the datacube indexed time. #45

### DIFF
--- a/cubedash/summary/_extents.py
+++ b/cubedash/summary/_extents.py
@@ -256,18 +256,50 @@ def _gis_point(doc, doc_offset):
     )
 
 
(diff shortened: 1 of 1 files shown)


jeremyh:
A rebuild of my local Africa copy worked, and the extents look like they're good now. Marking as ready.

alexgleith (approved):
Haven't tested, but looks good 👍

PR #197, opened 2020-09-30
Add condition for gha

pindge:
release `2.2.0` was created initially as a draft and triggered `created` condition for Github action, but it didn't contain a release tag, which meant the docker build didn't trigger.

- Adding second condition `published` for docker build
- `release tag` is faulty

https://github.com/opendatacube/datacube-explorer/runs/1186441298?check_suite_focus=true
```
Run echo $RELEASE
  echo $RELEASE
  shell: /bin/bash -e {0}
  env:
    OLD_IMAGE_NAME: opendatacube/dashboard
    IMAGE_NAME: opendatacube/explorer
    UNSTABLE_TAG: 2.2.0
    RELEASE: 2.1.12.dev166+ga34234b
2.1.12.dev166+ga34234b
``

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -9,7 +9,7 @@ on:
       - "**"
 
   release:
-    types: [created, published]
+    types: [created, edited, published]
 
 env:
   OLD_IMAGE_NAME: opendatacube/dashboard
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: .github/workflows/docker.yml [hunk 1]
```diff
@@ -9,8 +9,7 @@ on:
       - "**"
 
   release:
-    types:
-      - created
+    types: [created, published]
```
NikitaGandhi: can we add edited too?

NikitaGandhi (approved):

PR #199, opened 2020-10-07
Merge the `/about` and `/dashboard` pages

jeremyh:
Since the content is similar and the new dashboard is better.

- Reuse the table styling from the product-audit page. We'll probably make a new better global table style at some point.
- Add a csv download option since that's what Tish uses.
- Add license column for products. Currently this is only used by stac, but is mandatory in stac, so ideally we want people to start using it.

The new address is now `/about`, to replace the old about page which was just an unused product list. There's also `/about.csv` to get it in raw csv format.

Screenshot from 2020-10-07 11-19-44


jeremyh:
Open for other suggestions, of course :slightly_smiling_face:

whatnick (approved):

PR #200, opened 2020-10-08
Show example locations on the product page

jeremyh:
The new product table showed some location prefixes that raise immediate questions, such as "why is it only showing `file:/g/`?"

This adds example full-locations underneath on the products page to answer those questions:

Screenshot from 2020-10-08 11-06-33

(I've also moved on the product-related pages to the `_products.py` module rather than the global one)


whatnick (approved):
Moving forward

PR #203, opened 2020-10-19
Add Apache 2.0 License to datacube-explorer

woodcockr:
Hi Folks,

Not sure if this is the intended license but it would be consistent with the other Open Data Cube tools to have Explorer under the Apache 2.0 license.
So here's a PR to approve to make it so.


alexgleith (approved):
Thanks, @woodcockr ! Not sure how this was missed.

PR #209, opened 2020-11-05
Add geometry validity check

pindge:
**Goal:**
- stop invalid geometry from breaking explorer
- related issue: https://github.com/opendatacube/datacube-explorer/issues/146

**Scope:**
- explorer will only add dataset with footprint that is either `valid` or `null` to time overview
- all footprints together with its dataset info will be stored in dataset_spatial table

**Testing:**
- add s2_l2a product definition
- add one broken geometry and one normal geometry dataset
- test specific for s2_l2a scenario

**Testing setup**
- all dataset will be displayed under `/dataset/s2_l2a` page
- product overview `/s2_l2a` will only display dataset that has `valid` or `null` footprint


cubedash/summary/_extents.py
```diff
@@ -305,14 +306,22 @@ def refresh_product(
...
         ).rowcount
+
         log.debug(
             "extent_removal.end",
             deleted_count=change_count,
         )
+    if cleanup_invalid_geometry:
+        change_count += engine.execute(
+            DATASET_SPATIAL.delete().where(
+                func.ST_IsValid(DATASET_SPATIAL.c.footprint) is not True
```
alexgleith: I think this needs to be `!= true`, it's SQL and not Python.
```suggestion
                func.ST_IsValid(DATASET_SPATIAL.c.footprint) != True
```

cubedash/summary/_extents.py
```diff
@@ -384,6 +393,7 @@ def _populate_missing_dataset_extents(
                 == bindparam("product_ref", product.id, type_=SmallInteger)
             )
             .where(DATASET.c.archived == None)
+            .where(func.ST_IsValid(DATASET_SPATIAL.c.footprint) != True)
```
alexgleith: Not sure about this.

Ideally, the dataset is in Explorer, but the geometry is ignored? Or, we fix geometry at this step?

cubedash/summary/_extents.py
```diff
@@ -397,6 +407,7 @@ def _populate_missing_dataset_extents(
...
```
alexgleith: Same comment as above...

alexgleith (commented):
This looks good, but I'm not sure it's solving the problem in the right way at the right place...

cubedash/summary/_extents.py
whatnick: Use sqlalchemy construct instead : https://docs.sqlalchemy.org/en/14/core/sqlelement.html#sqlalchemy.sql.expression.false

whatnick (approved):
💯

PR #212, opened 2020-11-09
Fix SQLAlchemy breaking changes in version bump with pinned dependency

### LINKED ISSUE #211: The build works locally, but is hitting an error in pytest collection within Github Actions. @omad mentioned similar problems have been  happening elsewhere too.
The build works locally, but is hitting an error in pytest collection within Github Actions. @omad mentioned similar problems have been  happening elsewhere too.

```

==================================== ERRORS ====================================
________________________ ERROR collecting test session _________________________
/usr/lib/python3.6/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:994: in _gcd_import
    ???
<frozen importlib._bootstrap>:971: in _find_and_load
    ???
<frozen importlib._bootstrap>:955: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:665: in _load_unlocked
    ???
/env/lib/python3.6/site-packages/_pytest/assertion/rewrite.py:171: in exec_module
    exec(co, module.__dict__)
integration_tests/conftest.py:15: in <module>
    import cubedash
cubedash/__init__.py:6: in <module>
    from ._pages import app
cubedash/_pages.py:13: in <module>
    from cubedash import _audit, _monitoring
cubedash/_audit.py:14: in <module>
    from . import _model
cubedash/_model.py:17: in <module>
    from cubedash.summary import SummaryStore, TimePeriodOverview
cubedash/summary/__init__.py:3: in <module>
    from ._stores import SummaryStore
cubedash/summary/_stores.py:31: in <module>
    from sqlalchemy.engine import Engine, RowProxy
E   ImportError: cannot import name 'RowProxy'
```

_Originally posted by @jeremyh in https://github.com/opendatacube/datacube-explorer/pull/202#issuecomment-723680231_

whatnick:
Close #211 on successful CI run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,6 +88,9 @@ clean:  ## Clean all working/temporary files
 
 # DOCKER STUFF
 up: ## Start server using Docker
+	docker-compose up
+
+up-d: ## Start server using Docker in background
 	docker-compose up -d
 
 build: ## Build the dev Docker image
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ clean:  ## Clean all working/temporary files
 
 # DOCKER STUFF
 up: ## Start server using Docker
-	docker-compose up
+	docker-compose up -d
 
 build: ## Build the dev Docker image
 	docker-compose build
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -9,8 +9,6 @@ services:
     environment:
       - PROJ_LIB=/usr/share/proj  # This is to fix an odd issue with PyProj
       - prometheus_multiproc_dir=/tmp # Enable prometheus metrics
-      - DATACUBE_CONFIG_PATH=/root/.datacube_integration.conf # Enable test config
-      - DB_DATABASE=opendatacube_test
     volumes:
       - ./:/code
       - ./.docker/.datacube_integration.conf:/root/.datacube_integration.conf
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -9,6 +9,8 @@ services:
     environment:
       - PROJ_LIB=/usr/share/proj  # This is to fix an odd issue with PyProj
       - prometheus_multiproc_dir=/tmp # Enable prometheus metrics
+      - DATACUBE_CONFIG_PATH=/root/.datacube_integration.conf # Enable test config
+      - DB_DATABASE=opendatacube_test
     volumes:
       - ./:/code
       - ./.docker/.datacube_integration.conf:/root/.datacube_integration.conf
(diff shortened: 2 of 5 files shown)


jeremyh (approved):
Nice — thanks for looking into this 🙂

docker-compose.override.yml
```diff
@@ -9,6 +9,8 @@ services:
     environment:
       - PROJ_LIB=/usr/share/proj  # This is to fix an odd issue with PyProj
       - prometheus_multiproc_dir=/tmp # Enable prometheus metrics
+      - DATACUBE_CONFIG_PATH=/root/.datacube_integration.conf # Enable test config
+      - DB_DATABASE=opendatacube_test
```
alexgleith: This means it'll always use the test DB, right?

When you run the integration tests, it uses the `integration` config file, I believe. It's built in...

I don't think you need to make these two changes.

Makefile
```diff
@@ -88,7 +88,7 @@ clean:  ## Clean all working/temporary files
 
 # DOCKER STUFF
 up: ## Start server using Docker
```
alexgleith, line 90: For DEV, I like having Docker in the foreground. I'd prefer a different make command to run in the background.

Maybe we make an `up-d` command?

docker-compose.override.yml
pindge: no, this is to actually make this file useable for manual testing

alexgleith: I don't believe so, @pindge 

Datacube Core has a special case for config for running integration tests: https://github.com/opendatacube/datacube-core/blob/c2c25d3df437218223b900c1e6deb6f3dce374e4/README.rst#developer-setup

I don't think we necessarily want to use the `opendatacube_test` database for doing development work, which is what setting `DB_DATABASE` here will do.

alexgleith: @jeremyh  can confirm here 

But what you'll find is if you remove the volume mount and these two environment variables, the app will work fine. But integration tests will fail.

If you leave it the way I initially set it up, it works fine as a DEV install, and integration tests also work.

pindge: This file name is misleading, if its only to be used for integration time, it shouldn't be called `.override` should be `docker-compose.integration.yaml`

alexgleith: It's not meant only for integration time...

The override is the default, it's used for DEV and TEST.

In the Makefile, there's a `-prod` set of commands, that work without the override. That means you can test a build that doesn't have the code mounted, which is convenient for testing some things, like whether it gets installed right, and whether the version number is captured properly.

It is a bit complex, and could do with documenting! But it works, and I'm saying we don't need to change these aspects of it.

alexgleith (approved):

PR #216, opened 2020-11-12
cleanup dataset search page UI issues

pindge:
- style the select box to be 100% width not exceeding the page limit
- limit option description to 155 characters only so will not overflow a desktop display
- change `dataset-row-span` from 3 to 2 as the hardcoded `data-code-span` is only 1 + 1 = 2

related issue: https://github.com/opendatacube/datacube-explorer/issues/215


alexgleith (approved):

jeremyh (approved):
Looks good

PR #229, opened 2020-11-16
STAC fixes

alexgleith:
Implement a `numberMatched` and `context/matched` count of datasets that match the query.

Also fix a bug with queries with limit = 0.

Tests implemented ;-)

### DIFF
--- a/cubedash/summary/_stores.py
+++ b/cubedash/summary/_stores.py
@@ -1,10 +1,23 @@
+from dataclasses import dataclass
+
+import dateutil.parser
 import math
 import os
 import re
+import structlog
+from cachetools.func import ttl_cache
 from collections import Counter
-from dataclasses import dataclass
 from datetime import date, datetime, timedelta
+from dateutil import tz
+from geoalchemy2 import WKBElement
+from geoalchemy2 import shape as geo_shape
+from geoalchemy2.shape import to_shape
 from itertools import groupby
+from sqlalchemy import DDL, String, and_, func, select
+from sqlalchemy.dialects import postgresql as postgres
+from sqlalchemy.dialects.postgresql import TSTZRANGE
+from sqlalchemy.engine import Engine, RowProxy
+from sqlalchemy.sql import Select
 from typing import (
     Dict,
     Generator,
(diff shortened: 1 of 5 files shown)


jeremyh:
My reading was that `matched` was the total count of matching records in the database for the query, `returned` was the current page of results. (from [here](https://stacspec.org/STAC-api.html#operation/getFeatures)).

We're currently adding `matched` when we have a summary record that has a count already (where it's cheap and doesn't need a full db scan) https://github.com/opendatacube/datacube-explorer/blob/003cf56947749b02ed747ec4be50675bd8eacb2f/cubedash/_stac.py#L365-L368

jeremyh:
... though that's only for a full-collection query. In theory we could do it for others that match an existing summary.

To do it properly we'll need to add an extra `count()` query without a limit, to get the full, exact matching record count.

jeremyh (commented):
Looks nice! ... and you're quick too.

I think we need all of the `where` clauses to match, otherwise the counts wont match the results? It's missing the the `dataset_ids` and `requires_geometry` logic from `search_items`.

alexgleith:
> I think we need all of the `where` clauses to match, otherwise the counts wont match the results? It's missing the the `dataset_ids` and `requires_geometry` logic from `search_items`.

Yeah, you're right. Fixed, I think. I guess I should tweak the commits now...

cubedash/summary/_stores.py
```diff
@@ -815,6 +777,95 @@ def search_items(
...
+            dataset_ids=dataset_ids,
+            require_geometry=require_geometry,
+        )
+
+        result = self._engine.execute(query).fetchall()
+
+        if len(result) != 0:
+            return result[0][0]
+        else:
+            return 0
```
jeremyh, line 809: It's only cosmetic, but sqlalchemy has a neat `engine.execute(query).scalar()` function to query a single number succinctly.

jeremyh (approved):
:+1: looks good

cubedash/summary/_stores.py
alexgleith: Aha, thanks.

PR #230, opened 2020-11-17
Give correct match count when searching by ids

jeremyh:
And expand the test to cover it.

(otherwise a search such as `/stac/search?ids=cab65f3f-bb38-4605-9d6a-eff5ea786376` shows a huge count of all datasets in the datacube)


alexgleith (approved):
Oops, missed this one. Thanks @jeremyh!

PR #231, opened 2020-11-18
Add ''/arrivals" page to view recently added datasets

jeremyh:
Very simple, but useful right now. "What datasets have been added to my datacube most recently?"

It gives a count per-product and links to a few sample datasets:

Screenshot from 2020-11-18 13-42-16

It uses the existing  `dataset.created` index on ODC's table, so the query is pretty fast on my system. No caching added for now.


jeremyh:
It does not rely on any Explorer summaries either. So it's always up-to-date with the underlying datacube.

cubedash/_pages.py
```diff
@@ -35,6 +35,8 @@
 _HARD_SEARCH_LIMIT = app.config.get("CUBEDASH_HARD_SEARCH_LIMIT", 150)
 _DEFAULT_GROUP_NAME = app.config.get("CUBEDASH_DEFAULT_GROUP_NAME", "Other Products")
 
+_DEFAULT_ARRIVALS_DAYS: int = app.config.get("CUBEDASH_DEFAULT_ARRIVALS_DAY_COUNT", 14)
```
whatnick, line 38: Is there a way to make Flask app configs self-describing. There seem to be more and more knobs in explorer 💯

pindge (approved):

cubedash/_pages.py
jeremyh: Not that I've seen. The current method is to [mention them all in the README](https://github.com/opendatacube/datacube-explorer#how-can-i-configure-the-deployment)

( .... which I haven't done here. Give me 30 seconds :slightly_smiling_face:  )

jeremyh: Added in 8792a5a6fce05083b1be364431b5bca576df08e2

PR #235, opened 2020-11-26
Clean-up of page and URL hierarchies

jeremyh:
The addition of menus (#234) and breadcrumbs (#233) exposed how poorly structured our page hierarchy is. 

They both show a hierarchy of pages, and our own URLs show a hierarchy of pages, and all three are inconsistent.

This is a restructure to make the hierarchy more sane and explorable.

---

### Breadcrumbs

The current breadcrumb implementation only links to the _current_ page (and the 'home' page). All other breadcrumb elements are not clickable because there are no pages in the many parent URLs:

Screenshot from 2020-11-26 14-15-43

(I can't click "region" or the "product name". Only the link to the current page.)

This PR reorganises pages and their URLs enough so that every element of every breadcrumb is now a link that goes somewhere useful.

For instance, this is a page for a dataset:

Screenshot from 2020-11-26 14-18-55

- You can now click on every element: "products", "alos...", "datasets", and they all go to relevant parent pages.
- The url structure, grouped by product, roughly mirrors the Stac equivalent.
- All old URLs are maintained and contain redirects. So you can still link to `/dataset/<UUID>` and will be auto-redirected to the right collection.

### Top Menus
The top menu showed a different hierarchy and wasn't linking to every useful page in Explorer.

- This PR changes the top menu to match the URLs/Bread-Crumbs more closely and renames some pages to make more sense of their purpose.
- There's also more links to other pages that weren't in the top menu, like the product definitions and metadata types.

---

**Note:** There's still a bit of messiness here (especially the duplicate product lists), and I suspect we'll have some more PRs soon, but I figure it's better to iterate this gradually in smaller PRs.

**Note2:** I suspect we'll want to rename some of these URLs again tomorrow. Definitely expect iteration. Naming things is hard. We're not finalising any of these until we make a release.


jeremyh:
Broke every test imaginable. Converting to draft until fixed.

pindge (approved):

jeremyh:
This seems to be blocked by a Docker failure which is unrelated to to these changes.

jeremyh:
There's only test fixes since the review approval... so I'm going to bravely assume it's still approved.

PR #238, opened 2020-12-02
Add derived and source column to storage page

pindge:
- add derived and source products columns to storage page

This allows user to see at a glance products lineages, helps to determine when a product has 0 datasets and is not source product to another product it is a good indication this product is leftover of a deleted product set.

image

### DIFF
--- a/cubedash/templates/storage.html
+++ b/cubedash/templates/storage.html
@@ -19,7 +19,10 @@
         tbody tr:hover {
             background-color: #f0f0f0;
         }
-
+        li.lineage {
+            line-height: 15px !important;
+            list-style-type: none;
+        }
     </style>
 {% endblock %}
 {% block panel %}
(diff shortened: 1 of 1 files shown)


jeremyh (approved):
:+1: 

Stretch goal some day: draw the product connections as a pretty graph

jeremyh:
Note that those source/derived links are an estimate, from sampling a random % of dataset provenance in each product.

If the list needs to be exactly-correct, we'd need to make that query a little more expensive.

whatnick:
> 
> 
> 👍
> 
> Stretch goal some day: draw the product connections as a pretty graph

BlockChain of products / datasets ?

whatnick (approved):
💯

pindge:
> > 👍
> > Stretch goal some day: draw the product connections as a pretty graph
> 
> BlockChain of products / datasets ?

sounds like something i fancy doing

PR #240, opened 2020-12-02
Add a page explaining each of Explorer's APIs

jeremyh:
The current links to raw APIs in the menus are quite confusing, as clicking them gives the user raw json.

Move the APIs to a new page that explains each one:

Screenshot from 2020-12-03 10-09-55


pindge (approved):
Finally about page doing what about is about

Issue #241, opened 2020-12-02
'datetime' STAC searches don't support all of RFC 3339


jeremyh:
We allow stac searches with a datetime property: `/stac/search?datetime=`

Currently `datetime` can be a timestamp:
 ```
/stac/search?datetime=2018-02-12T23:20:50Z
```

or a range:

```
/stac/search?datetime=2018-02-12T00:00:00Z/2018-03-18T12:31:12Z
```
... and time is optional:

```
/stac/search?datetime=2018-02-12
/stac/search?datetime=2018-02-12/2018-03-01
```

But [RFC 3339](https://tools.ietf.org/html/rfc3339) looks much larger and has many other options. Most notably, newer version of stac-api mention that open ranges are allowed with two dots:

```
/stac/search?datetime=2018-02-12/...
```

Short-term, we should at least support these open-ranges, since the [Stac docs](https://stacspec.org/STAC-api.html#operation/getFeatures) mention them, and they're easy to implement.

...longer-term, ideally look for a library to parse these ranges rather than implementing them ourselves.

jeremyh:
Follow-up query to look into: do our range searches handle inclusive/exclusive bounds the same way as expected by Stac?

Do we interpret "ranges" like `/stac/search?datetime=2018-02-12/2018-02-12` correctly? (I don't know what "correct" is without research)

jeremyh:
See also comment by @alexgleith on limitation of current open-range searches: https://github.com/opendatacube/datacube-explorer/pull/243#discussion_r534572906

PR #243, opened 2020-12-03
Support open time ranges in Stac searches

jeremyh:
For #241

Example:
```
/stac/search?datetime=2018-02-12/..
```


cubedash/_stac.py
```diff
@@ -423,18 +423,40 @@ def _pick_remote_uri(uris: Sequence[str]) -> Optional[int]:
...
+    >>>
     """
     time_period = time.split("/")
     if len(time_period) == 2:
-        return parse_time(time_period[0]), parse_time(time_period[1])
+        start, end = time_period
+        if start == "..":
+            start = datetime(1971, 1, 1, 0, 0)
+        elif end == "..":
+            end = datetime.now() + timedelta(days=2)
```
alexgleith, line 454: I guess we don't have data in the future, but this is an assumption. What about forecast data?

cubedash/_stac.py
```diff
@@ -423,18 +423,40 @@ def _pick_remote_uri(uris: Sequence[str]) -> Optional[int]:
...
+        # Were they both open? Treat it as no date filter.
+        if end == "..":
```
alexgleith, line 456: Should this be an `else`?

OR maybe it should be `if start == ".." and end == ".."`. Make it explicit?

alexgleith (approved):
Couple of comments, but looks good.

cubedash/_stac.py
jeremyh: It works because the second branch is an “elif” instead of an “if”. I was trying to be clever and reduce comparisons for some reason.

jeremyh: But yeah, clarity is better...

alexgleith: Oh I see. By this point it's not `..` if the start has a value and the end doesn't! It's obvious now.

cubedash/_stac.py
jeremyh: (missed this comment on my mobile sorry)

... I’m not sure! I didn’t know that was a thing.

Perhaps it could be set to `max(now(), most_recent_dataset_time)`, though that’ll need a query (whose result could be cached fairly decently).

jeremyh: Postgres has open ranges too, which is a better solution. We don’t currently expose them in the lower api.

jeremyh: Or simplest solution: choose an arbitrary long-time-in-the-future that we don’t expect to cause date bugs.

Issue #227, opened 2020-11-16
STAC API doesn't have Assets for some products


alexgleith:
For example, `ga_ls_wo_3` doesn't have assets: https://explorer.dev.dea.ga.gov.au/stac/collections/ga_ls_wo_3/items

Also, `ga_ls8c_ard_3` doesn't have assets: https://explorer.dev.dea.ga.gov.au/stac/collections/ga_ls8c_ard_3/items

Note that assets need to have `proj` fields to be able to be indexed easily into the ODC:

```
"B03": {
"title": "Band 3 (green)",
"type": "image/tiff; application=geotiff; profile=cloud-optimized",
"roles": [
"data"
],
"gsd": 10,
"eo:bands": [
{
"name": "B03",
"common_name": "green",
"center_wavelength": 0.56,
"full_width_half_max": 0.045
}
],
"href": "https://sentinel-cogs.s3.us-west-2.amazonaws.com/sentinel-s2-l2a-cogs/18/T/UM/2020/11/S2A_18TUM_20201116_0_L2A/B03.tif",
"proj:shape": [
10980,
10980
],
"proj:transform": [
10,
0,
300000,
0,
-10,
4700040,
0,
0,
1
]
},
```

jeremyh:
They are currently excluded by default unless you set `STAC_DEFAULT_FULL_ITEM_INFORMATION = True` in the deployment config (see [doc](https://github.com/opendatacube/datacube-explorer#how-can-i-configure-the-deployment) and [pr description](https://github.com/opendatacube/datacube-explorer/pull/202)) 

... or individually by adding `&_full=true` to the query URL, eg: `https://explorer.dev.dea.ga.gov.au/stac/collections/ga_ls_wo_3/items?_full=true`

It's turned off by default in the current release because the "full information" requires going to ODC tables (instead of just Explorer tables), which are extremely slow in the current implementation, especially with paging, and cause timeouts in too many searches in our cloud deployment.

.... but the full asset details are always accessible if you follow the url for each search result, eg: https://explorer.dev.dea.ga.gov.au/stac/collections/ga_ls_wo_3/items/03feb8ce-b371-5b4f-9097-90c9e4051ea5

It's not ideal...  but timeouts mean that no one can fetch anything. For now it's opt-in to a slower api.

PR #250, opened 2020-12-17
Enable full Item Information on Stac searches

jeremyh:
Add an index that speeds up our default Stac Item searches.

... and change the Stac API to always return "Full Item Information" by default (addresses #227)

---

The new index removes the need for the DB to re-sort all items for paging on every query, so it is significantly faster (...at least in my clone-of-nci tests).

Users can rerun `cubedash-gen -v --init` to add the "missing" index. 

 - *Note* This addition takes an exclusive lock, but only runs for a few seconds on my local full copy of the NCI index (our biggest index). 
     - It's only running on the (relatively small) spatial table.
 - There's no forward or backward incompatibilities: the index is optional for current and past versions of Explorer, so can be run at any time. Or not run. I've not flagged it as an incompatible schema for Explorer.

---
Misc
- Ensure all api endpoints give 404 codes, not exceptions. I hit this during testing.
- Removed some of the older, defunct code for asset handling. It has been replaced by eodatasets3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ You can alter default [Flask](http://flask.pocoo.org/docs/1.0/config/) or
     # Full searches are much slower because they use ODC's own raw metadata table.
     # (Users can append "_full=true" to requests to manually ask for full metadata.
     #  Or preferrably, follow the `self` link of the Item record to get the whole record)
-    STAC_DEFAULT_FULL_ITEM_INFORMATION = False
+    STAC_DEFAULT_FULL_ITEM_INFORMATION = True
 
 
 [Sentry](https://sentry.io/) error reporting is supported by adding a `SENTRY_CONFIG` section.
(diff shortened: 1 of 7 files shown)


jeremyh:
+1.5% 🙂

jeremyh:
@whatnick We'll want to run `cubedash-gen -v --init` in our environments once this is in. Though unlike the previous change it can be run at any time (before or after deployment), and is quite cheap (no need to recreate summaries or anything).

README.md
```diff
@@ -179,7 +179,7 @@ You can alter default [Flask](http://flask.pocoo.org/docs/1.0/config/) or
     # Full searches are much slower because they use ODC's own raw metadata table.
     # (Users can append "_full=true" to requests to manually ask for full metadata.
     #  Or preferrably, follow the `self` link of the Item record to get the whole record)
-    STAC_DEFAULT_FULL_ITEM_INFORMATION = False
+    STAC_DEFAULT_FULL_ITEM_INFORMATION = True
```
alexgleith, line 182: :-)

alexgleith (approved):
Looks good to me!

PR #264, opened 2021-03-01
Avoid duplicating linting settings

jeremyh:
The two differ _slightly_ at the moment. We should just ask pre-commit to run its checks.


jeremyh:
PR #265 will need to be merged to fix the build

whatnick (approved):

Issue #201, opened 2020-10-08
The /audit/storage page is slow to load on dev


jeremyh:
My guess is that this is due to the number of product-location queries.  They're done for every row of the table, so it's an NxM query situation. On my local machine, with near-zero latency to local database per-query, it's quite fast to load. 

<sup>(this is all a guess, of course we should measure)</sup>

Those calculations could be cached heavily as they're the same for everybody, don't change often, and are only estimates anyway (based on a tiny sample).

jeremyh:
Alternative option: pre-calculate them ahead of time in summary generation. It could then be an exact answer, rather than an estimate.

omad:
Between #268 and increasing the cache duration, this is probably no longer an issue. Closing.

PR #268, opened 2021-03-10
Radically reduce query count on storage page

jeremyh:
Make it fetch the whole table in one query, instead of N queries for N rows of the table.

Doing N queries is extremely slow in our cloud instances, as the higher database latency is repeated many times, so the page would take a _long_ time to load.

I believe this change should address #201

---

I'm also bumping the default cache TTL to be higher, as none of our environments change fast enough for sub-minute accuracy to matter.


uchchwhash (approved):
Awesome!

alexgleith (approved):
Nice work!

PR #269, opened 2021-03-11
Expand incremental-update function: no fast-product-skipping, allow marker reset

jeremyh:
Most importantly, **tldr**: this fixes an issue where if `generate` is killed half way through processing, you _may_ miss some new datasets in the statistics.

### Removal of fast-product-skipping
The `generate` command would skip month-and-year recalculations when no datasets seemed to have changed.

... but our way of calculating dataset changes can be inaccurate if `generate` was killed during its previous execution. Which would make us skip the product incorrectly.

Instead, we should still check that all months and years are accurate each time we run.  This slightly slows down incremental updates, but they're still very quick (seconds) locally in my testing. Which is better than being corruptible via system crashes.

### Add `--reset-incremental-position` to the generate command

This option will make the incremental-updater more conservative in where it starts scanning from: it moves the marker back to the last position with a newly-added dataset in Explorer's own tables. 

This is useful in development when cloning an external DB, as we want to reset the incremental marker to fit the new DB. 

(The catch: it makes the update take longer as more data than strictly necessary will be recomputed.)

### Cleaner logging, user messages.

.... and say "No changes", rather than "Skipped", since that's more accurate.

### DIFF
--- a/cubedash/generate.py
+++ b/cubedash/generate.py
@@ -66,8 +66,12 @@ from click import secho as click_secho
 from click import style
 
 from cubedash.logs import init_logging
-from cubedash.summary import SummaryStore, TimePeriodOverview
-from cubedash.summary._stores import GenerateResult
+from cubedash.summary import (
+    SummaryStore,
+    TimePeriodOverview,
+    GenerateResult,
+    UnsupportedWKTProductCRS,
+)
 from datacube.config import LocalConfig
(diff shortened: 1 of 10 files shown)


jeremyh:
Still a draft as I'm going to do another clone/migration run with this code overnight to look for any other issues.

uchchwhash (approved):
Cool!

Issue #270, opened 2021-03-16
Feature Request : Publish Explorer via PyPI


whatnick:
Explorer already has a clearly defined set of production, testing and development dependencies in setup.py. Leverage this list to publish it as python wheel / source package on PyPI.

jeremyh:
I just tried doing this, but it looks like it's blocked by our usage of a direct (Git) dependency on `flask_themes`: https://github.com/opendatacube/datacube-explorer/blob/90c131368121af040d63618b426d031092880cee/setup.py#L65

jeremyh:
Perhaps investigate moving to something more maintained, like [flask-themer](https://pypi.org/project/flask-themer/)* 

*this is not an endorsement; it was just the first okay-sounding one on a Google search

jeremyh:
Now installable from pypi :tada:  https://github.com/opendatacube/datacube-explorer/releases/tag/2.6.0

PR #271, opened 2021-03-23
Prepare for future release to Pypi

jeremyh:
We have one dependency that blocks us from releasing to pypi just yet, but there are enough improvements in this PR to merge as-is.

As requested in #270

- [X] Test that it works from a wheel install (are all static files included properly?)
- [X] Add a command to run Explorer directly (the `run.sh`/`run-dev.sh` scripts wont be available from pypi)
           (**added `cubedash-run` command**)
- [X] Update readme to include the simpler ~pypi setup~ pip setup (until last element resolved)
- [x] Remove direct (Git-based) [dependency](https://github.com/opendatacube/datacube-explorer/blob/90c131368121af040d63618b426d031092880cee/setup.py#L65) on flask-themes, replace with more modern  [flask-themer](https://pypi.org/project/flask-themer/) (which is available as a dependency on pypi)
- [x] ~Wait for next flask-themer release, to fix the static file loading problem.~ (leave for future PR)


whatnick:
The direct dependency could be removed by publishing another PyPI package, or checking PyPI for something else.

jeremyh:
Yeah, I was looking at flask-themer potentially as a replacement:  https://pypi.org/project/flask-themer/

jeremyh:
Ok, switched to flask-themer... and it too has a (completely different) bug that prevents us from using their pypi release.

(They're overriding the `/static` endpoint for non-theme templates. Fixed in October. I asked them if they can make a release)

uchchwhash (approved):
💯

alexgleith (approved):
Awesome!

PR #276, opened 2021-04-06
Add CORS headers to support web-based STAC browsers

andrewdhicks:
Web-based STAC browsers such as https://rocket.snapplanet.io/home can't load the STAC catalog provided by Explorer due to Cross-Origin Resource Sharing restrictions.

Simple implementation for the entire app using https://flask-cors.readthedocs.io/ by setting config variable
`CUBEDASH_CORS = True`

Should CORS only be added to STAC-endpoints, or limited to specific domains? 
eg `cors = CORS(app, resources={r"/stac/*": {"origins": "*"}})`


jeremyh:
Sounds like a good idea to me. `/stac` and `/api` are the obvious endpoints to include.

Given that Explorer is read-only, with no authentication or operations to perform, I'm wondering whether it would hurt to default it to "on".

andrewdhicks:
@jeremyh I've added `/stac/` and `/api/` as the list of routes to use CORS, and defaulted CORS to be on.
I don't have a working ODC environment, so I haven't test that any of this works yet...

jeremyh (approved):
It looks like it works on my local install:

Screenshot from 2021-04-08 16-12-29

Thanks Andrew!

PR #280, opened 2021-05-05
Set up data links on dataset page

alexgleith:
* Adds a list of data URIs to download TIFs

Jeremy, I'm sure I'm missing something. Let me know what I can do to make this more robust?

### DIFF
--- a/cubedash/summary/_stores.py
+++ b/cubedash/summary/_stores.py
@@ -34,14 +34,14 @@ from typing import (
     Tuple,
     Union,
 )
-from urllib.parse import urljoin
+from urllib.parse import urljoin, urlparse
 from uuid import UUID
 
 try:
     from .._version import version as explorer_version
 except ModuleNotFoundError:
     explorer_version = "ci-test-pipeline"
-from cubedash import _utils
+from cubedash import _model, _utils
 from cubedash._utils import ODC_DATASET, ODC_DATASET_LOCATION, ODC_DATASET_TYPE
 from cubedash.summary import RegionInfo, TimePeriodOverview, _extents, _schema
 from cubedash.summary._extents import (
@@ -1574,9 +1574,7 @@ class SummaryStore:
 
         if is_doc_eo3(dataset.metadata_doc):
             dataset_doc = serialise.from_doc(dataset.metadata_doc, skip_validation=True)
-            dataset_location = (
-                dataset_doc.locations[0] if dataset_doc.locations else None
-            )
+            dataset_location = dataset.uris[0] if dataset.uris else None
 
             uri_list = {
                 name: urljoin(dataset_location, m.path)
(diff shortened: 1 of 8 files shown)


cubedash/summary/_stores.py
```diff
@@ -1561,6 +1563,21 @@ def get_dataset_footprint_region(self, dataset_id):
             row.region_code,
         )
 
+    def get_dataset_uris(self, dataset: Dataset):
+        """
+        Get a representation of the data URIs so that they
+        can be linked to to download them.
+        """
+        dataset_doc = serialise.from_doc(dataset.metadata_doc, skip_validation=True)
```
jeremyh: This parser only reads eo3 docs, which is why the tests of older datasets are failing.

Simple option may be to check if it's eo3 with [eodatasets3.utils.is_doc_eo3(doc)](https://github.com/GeoscienceAustralia/eo-datasets/blob/8f5088ce0281b9b4b465fcbfb71fdd58d3c70a83/eodatasets3/utils.py#L144-L145) and not show the list otherwise.

cubedash/summary/_stores.py
```diff
@@ -1561,6 +1563,21 @@ def get_dataset_footprint_region(self, dataset_id):
...
+        dataset_location = dataset_doc.locations[0] if dataset_doc.locations else None
+
+        uri_list = {
+            name: urljoin(dataset_location, m.path)
+            for name, m in dataset_doc.measurements.items()
```
jeremyh: We probably want to include `dataset_doc.accessories.items()` too. 

They are all the dataset files that aren't measurements: non-odc metadata, thumbnails, checksums, etc.

jeremyh (commented):
Nice idea!

cubedash/summary/_stores.py
jeremyh: If we want to support everything, a more-thorough option could be to extract the basic [eo-to-eo3 logic](https://github.com/opendatacube/datacube-explorer/blob/ee63dcb381e623dba87c07375d5b592eea952ec2/cubedash/_stac.py#L638-L640) into a function, and run that first. (but ideally we might move it into eodatasets proper)

cubedash/summary/_stores.py
alexgleith: Done

cubedash/summary/_stores.py
```diff
@@ -1561,6 +1564,34 @@ def get_dataset_footprint_region(self, dataset_id):
...
+        can be linked to to download them.
+        """
+
+        uri_list = {}
+
+        if is_doc_eo3(dataset.metadata_doc):
+            dataset_doc = serialise.from_doc(dataset.metadata_doc, skip_validation=True)
+            dataset_location = (
+                dataset_doc.locations[0] if dataset_doc.locations else None
+            )
```
jeremyh: When running everything's coming back as a relative path (which are broken as they're relative to Explorer's webpage url). Presumably it's because this field is null.

(The raw indexed `dataset.metadata_doc` column doesn't contain locations)

We'll need to grab locations from the original Dataset object instead:
```suggestion
            dataset_location = (
                dataset.uris[0] if dataset.uris else None
            )
```

cubedash/templates/dataset.html
```diff
@@ -4,6 +4,8 @@
 
 {% set extent_geojson = dataset_footprint.__geo_interface__ %}
 
+{% set data_uris = data_uris %}
```
jeremyh: Does that do something in jinja?

jeremyh (commented):
The paths are quirky for me when running locally -- see below.

(I'd also like to fiddle with the css/styles of the panel at some point if I get the time, to make it more consistent with others... but it wouldn't block this pr.)

cubedash/templates/dataset.html
alexgleith: Oh, probably not. I was just copying the patterns from before it!

PR #282, opened 2021-05-20
Fix dependency issues and update Dockerfile

alexgleith:
* Fix issue with Jinja imports changing
* Redo the Dockerfile to update to the new geobase pattern
* Fix some dependencies, remove others, simplify.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -58,7 +58,7 @@ setup(
         "click",
         "dataclasses>=0.6;python_version<'3.7'",
         "datacube>=1.8",
-        "eodatasets3 >= 0.17.0",
+        "eodatasets3>=0.17.0",
         "fiona",
         "flask",
         "Flask-Caching",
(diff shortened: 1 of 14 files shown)


setup.py
```diff
@@ -57,8 +57,8 @@
         "cachetools",
         "click",
         "dataclasses>=0.6;python_version<'3.7'",
-        "datacube>=1.8",
-        "eodatasets3 >= 0.17.0",
+        "datacube",
+        "eodatasets3",
```
jeremyh: These are still valid, aren't they? They're essential minimums

jeremyh: They'll still be relevant in the setup.py for regular pip installations (... and for when it goes to pypi)

alexgleith: Hmm, I guess so. I get confused as to when to include in setup.py vs requirements.txt vs constraints.txt still! I'll put them back.

jeremyh (approved):
Thanks for fixing these! 🙂

PR #285, opened 2021-06-02
add curlylint for jinja template

pindge:
- setup curlylint to ensure jinja templates are correct https://www.curlylint.org/


jeremyh:
Neat! I love linting :slightly_smiling_face: 

All of our linting is inside `.pre-commit-config.yaml`, which is ideal because it's run everywhere already (before committing, in Github Actions, via `make lint` etc), so ideally we should include it there?  Otherwise people wont discover their errors until after pushing. And maintaining it should be simpler than managing a custom installation in Github Actions.

It looks like they have pre-commit support: https://www.curlylint.org/docs/reference/integrations#usage-with-pre-commit-git-hooks-framework

jeremyh (approved):

PR #287, opened 2021-06-17
Fix up as_external_url to do correct S3 HTTPS URI

alexgleith:
My favourite PR, I think. Change `-` to `.` :-) 

Justification is this doesn't work: https://deafrica-services.s3-af-south-1.amazonaws.com/ga_ls8c_wofs_2_summary/WOFS_6933_25_5_summary_count_clear.tif

This does: https://deafrica-services.s3.af-south-1.amazonaws.com/ga_ls8c_wofs_2_summary/WOFS_6933_25_5_summary_count_clear.tif

### DIFF
--- a/cubedash/_utils.py
+++ b/cubedash/_utils.py
@@ -133,7 +133,7 @@ def as_external_url(url: str, s3_region: str = None) -> Optional[str]:
     parsed = urlparse(url)
 
     if s3_region and parsed.scheme == "s3":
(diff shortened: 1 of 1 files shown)


jeremyh:
Cool. There’s a doc test for it that’ll need to be updated for the tests to pass: https://github.com/opendatacube/datacube-explorer/blob/develop/cubedash/_utils.py#L124-L125

(… but it also looks like a dependency update has broken other tests outside of this PR. Might look at freezing our dependencies at some point.)

jeremyh (approved):

PR #298, opened 2021-07-09
Add GeoJSON and All-Products download options

jeremyh:
## Yaml downloads

Add paths to download all products/metadata-types as a single yaml file:

- All Product definitions: `http://explorer.example/products.odc-product.yaml`
- All Metadata Type definitions: `http://explorer.example/metadata-types.odc-type.yaml`

So people can, for example, clone remote datacube config in two commands:
```
datacube type add http://explorer.test/metadata-types.odc-type.yaml
datacube product add http://explorer.test/products.odc-product.yaml
```

## GeoJSON downloads

Add links in the sidebar to download the displayed product extents as GeoJSON files:

Screenshot from 2021-07-09 11-31-35
(People have asked for these on Slack, as well as for DEA's product coverage as a shape)

### DIFF
--- a/cubedash/_api.py
+++ b/cubedash/_api.py
@@ -1,5 +1,6 @@
 import logging
 
+import flask
 from flask import Blueprint, abort, request
 
 from cubedash import _utils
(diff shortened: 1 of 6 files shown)


alexgleith (approved):
I like it!

PR #299, opened 2021-07-09
Extra yaml download links, names

jeremyh:
Tiny amendments to #298 

- Add the Explorer endpoint name to the download filename hint. 
    - `products-deafrica.odc-product.yaml`, rather than `products.odc-product.yaml`. 
    - (Helpful when you have multiple in your downloads folder.)
- Include Yaml download icons on the products/metadata-types pages:

Screenshot from 2021-07-09 14-54-24

Screenshot from 2021-07-09 14-54-44


uchchwhash (approved):
👍🏽

PR #307, opened 2021-08-16
Bump Stac support to 1.0.0

jeremyh:
Updates the existing Stac validation methods for 1.0.0 (as there's a script to do so). I know we ideally want to move to pystac or similar eventually.

~Requires the develop version of eodatasets for Stac 1.0.0 conversion, which is not yet in stable, so it will not build successfully yet in our CI.~ released


jeremyh:
Also requires https://github.com/GeoscienceAustralia/eo-datasets/pull/202

pindge (approved):

Issue #310, opened 2021-08-25
Add region_code region information to the pages


alexgleith:
Pages like [this](https://explorer.digitalearth.africa/product/s2_l2a/regions/38PLQ) display the `region_code`, but don't have any information on it.

If there was a clickable link to the GeoJSON, that would be nice! Or a list of BBOX coords, perhaps.

![image](https://user-images.githubusercontent.com/3445853/130878046-d3cc8a97-e228-41ee-a3c3-274cd02fbc03.png)

pindge:
This has been implemented

![image](https://user-images.githubusercontent.com/24644475/173940848-420a0c37-2147-426f-810c-332a359d63eb.png)

PR #312, opened 2021-09-01
Add GeoJSON button+endpoint to region page

jeremyh:
As suggested in #310:

Screenshot from 2021-09-01 15-55-55


uchchwhash (approved):
thanks @jeremyh!

alexgleith (commented):
Love your work!

Issue #320, opened 2021-09-20
STAC broken in latest release


benjimin:
Explorer is supposed to have some support for STAC queries. Apparently this worked previously but is broken in the most recent release.

Using: `pystac-client`:
```
$ stac-client collections http://explorer.dea.ga.gov.au/stac

JSON does not represent a STAC object.
```

Using `odc-tools`:
```
$ stac-to-dc --catalog-href=https://explorer.dea.ga.gov.au/stac/ --collections=wofs_albers --bbox=35,-26,36,-25 --limit=10

pystac.errors.STACTypeError: JSON does not represent a STAC object.
```

According to @alexgleith this worked last week but not now:
```
stac-to-dc --catalog-href=https://explorer.digitalearth.africa/stac/ --collections=s1_rtc --bbox=35,-26,36,-25 --limit=10
```

benjimin:
Related: would be great to have more stac tests (#258 & #304) to catch this kind of thing.

Possibly related to latest stac release (#303)? (Or a nonstandard implementation confusing the client library? #181)

alexgleith:
Full error is: `stac-to-dc --catalog-href=https://explorer.digitalearth.africa/stac/   --collections=s1_rtc --bbox=35,-26,36,-25 --limit=10`

``` 
Traceback (most recent call last):
  File "/usr/local/bin/stac-to-dc", line 33, in <module>
    sys.exit(load_entry_point('odc-apps-dc-tools', 'console_scripts', 'stac-to-dc')())
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/Users/alex/git/opendatacube/odc-tools/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py", line 250, in cli
    added, failed = stac_api_to_odc(
  File "/Users/alex/git/opendatacube/odc-tools/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py", line 125, in stac_api_to_odc
[...]

PR #321, opened 2021-09-20
Enable validation on root stac catalog

jeremyh:
... and fix the missing field (needed for 1.0.0)!

For #320


alexgleith (approved):

jeremyh:
Added missing validation for '/arrivals' catalog too.

PR #324, opened 2021-10-05
Move Docker image to the standard gdal base

jeremyh:
It's less maintenance than the geo-base one, and brings us in-line with datacube-alchemist and eodatasets.

It also fixes the current security-scan failure.

### DIFF
--- a/constraints-docker.txt
+++ b/constraints-docker.txt
@@ -1,4 +1,1 @@
-datacube==1.8.3
-sqlalchemy==1.3.20
-click<8.0.0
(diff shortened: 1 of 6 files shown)


constraints-docker.txt
```diff
@@ -1,4 +1 @@
-datacube==1.8.3
-sqlalchemy==1.3.20
-click<8.0.0
-eodatasets3>=0.17.0
```
jeremyh, line 4: There's appropriate minimum versions in the dependencies now, I believe. So I removed these.

PR #328, opened 2021-10-12
Make grouping CRS configurable

### LINKED ISSUE #206: Footprint CRS is hardcoded
This really shouldn't be a region-centric CRS, as footprints away from Australia are likely to be quite distorted.

https://github.com/opendatacube/datacube-explorer/blob/17d840aa35d4fdd11a594777589cd2a29cc86444/cubedash/summary/_schema.py#L40

jeremyh:
Closes #206 

- A different CRS can be chosen on init of your Explorer schema: `cubedash-gen --init --epsg 3577`
- If not specified, the default is now global equal-area `epsg:6933` instead of Australian Albers.

### DIFF
--- a/cubedash/generate.py
+++ b/cubedash/generate.py
@@ -77,6 +77,7 @@ from cubedash.summary import (
     TimePeriodOverview,
     UnsupportedWKTProductCRS,
 )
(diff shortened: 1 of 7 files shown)


alexgleith (approved):
I think this is a GREAT change!

PR #329, opened 2021-10-12
Update default Docker command to match deployment

jeremyh:
Someone on Slack pointed out the default Docker CMD is not working, as meinheld has a dependency conflict.

We don't use meinheld on our prod deployments, so didn't notice.


alexgleith (approved):
Nice!

Issue #337, opened 2021-11-04
Instability and ramping resource usage


benjimin:
Lately have been observing some instability in this explorer deployment.

There is sawtooth **ramping** of the CPU utilisation, memory utilisation, and transmit bandwidth. That is, each new pod steadily increases both its CPU utilisation and its transmit bandwidth over the course of roughly 24 hours, and then the pods are collectively replaced. The stats peak at roughly 75% CPU utilisation, and 5MB/s egress per pod. The receive bandwidth is _not_ ramping (maintaining a fairly consistent 0.1MB/s per pod).

![image](https://user-images.githubusercontent.com/12852539/140258446-63123671-1a58-4a29-9185-d207a15b04f2.png)

Also (perhaps unrelated) there have been frequent failures from the front page (most commonly 502, 503, and timeouts) according to monitoring by uptime robot.

benjimin:
One evident problem is an unbounded set of labels applied to prometheus metrics. (As noted [in a recent datacube-wps issue](https://github.com/opendatacube/datacube-wps/issues/129#issuecomment-950014759), metrics are not for logs.)

```
flask_http_request_duration_seconds_sum{method="GET",path="/product/high_tide_comp_20p/regions/6_-41",status="404"} 0.052539687836542726
flask_http_request_duration_seconds_sum{method="GET",path="/products/wofs_annual_summary/datasets/17c40f41-65d4-4e43-968b-5cf5d953fed6",status="200"} 0.1782301519997418
flask_http_request_duration_seconds_sum{method="GET",path="/dataset/0dcdef87-b507-42dc-a69b-c570501068ca.odc-metadata.yaml",status="200"} 0.031694565899670124
flask_http_request_duration_seconds_sum{method="GET",path="/products/ls7_fc_albers/datasets/6b772d4a-7c00-46b9-8b1a-38e4d95c3380",status="200"} 0.14692931901663542
flask_http_request_duration_seconds_sum{method="GET",path="/products/ga_ls5t_ard_3/datasets/379642d8-6ebe-4b55-9c71-92e345567dd5",status="200"} 0.11339368112385273
flask_http_request_duration_seconds_sum{method="GET",path="/dataset/65f34855-7322-4cf2-9a93-cffd12639cf7.odc-metadata.yaml",status="200"} 0.017415384063497186
[...]

PR #338, opened 2021-11-09
Fix for stability/metrics

benjimin:
* Address #337 (explosion of metrics) by labelling metrics by endpoint, rather than by path. Because the number of distinct paths is nearly unbounded (distinguishing those is a role for logs not metrics).
* Resolve repo staleness (previously passing tests now fail) by pinning a dependency version to the last compatible version (pending further updates).

### DIFF
--- a/constraints-docker.txt
+++ b/constraints-docker.txt
@@ -1,1 +1,2 @@
 datacube==1.8.6
+eodatasets3==0.22.2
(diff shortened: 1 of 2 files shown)


jeremyh (approved):
Thanks for looking into this

jeremyh:
eodatasets regression is fixed in https://github.com/opendatacube/eo-datasets/pull/220

PR #339, opened 2021-11-10
Capitalise prometheus env var

benjimin:
* Capitalise the Prometheus environment variable name
  * By convention, environment variable names should always be uppercase.
  * Using lowercase is now triggering a _deprecation warning_ (visible in CI test runs).
  * We have seen bugs (e.g. [opendatacube/datacube-wps#129](https://github.com/opendatacube/datacube-wps/issues/129)) related to continued use of lowercase.
* Default to explicitly using the shared memory location for Prometheus (aggregating metrics data between processes).
   * We want to ensure we're using a `tmpfs` mount, to avoid potential IO performance issues. 
   * On the current container images, `/tmp` is not `tmpfs`.

### DIFF
--- a/cubedash/_model.py
+++ b/cubedash/_model.py
@@ -287,7 +287,7 @@ def enable_sentry():
 @app.before_first_request
 def enable_prometheus():
     # Enable deployment specific code for Prometheus metrics
-    if os.environ.get("prometheus_multiproc_dir", False):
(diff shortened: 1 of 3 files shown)


jeremyh (approved):

benjimin:
@jeremyh this shouldn't directly relate to #337, but we should coordinate the release with a corresponding helm chart update (to avoid a gap in monitoring).

PR #343, opened 2021-11-22
update docker-composer postgis image tag to 12-3.1

NikitaGandhi:
We are looking to upgrade AWS aurora postgres version to 13.3.3 and postGIS version to 3.1 so making sure ows and explorer has same checks


jeremyh (approved):
👍

PR #344, opened 2021-11-24
Support datasets with null datetimes

jeremyh:
(Fixes summary generation failure in Africa.)

In newer versions of Stac, `datetime` became optional, in which case the start/end time must be specified. When it's converted to eo3 on index we don't do anything about this, so effectively that's true of eo3 too now.

Add support for this case, with an example test dataset from Africa.


NikitaGandhi (approved):
Thanks @jeremyh!

alexgleith (approved):

Issue #279, opened 2021-05-05
"About Products" page is missing a / in the s3:// protocol


alexgleith:
Pages like this: https://explorer.dev.digitalearth.africa/audit/storage

Have a location like this: `s3:/deafrica-input-datasets/alos_palsar_mosaic/2018`

I think it should be like this `s3://deafrica-input-datasets/alos_palsar_mosaic/2018`

jeremyh:
Fixed in 2.8.1

PR #349, opened 2021-12-07
Fix the missing slashes in locations

jeremyh:
Our common location calculation has missing slashes:
Screenshot from 2021-12-08 10-17-26

This seems to be the behaviour of Python's `os.path.commonpath`, which was was written for paths, not URLs.

I've added a `common_uri` function, and add regression tests.


jeremyh:
Reported in #279

PR #374, opened 2022-05-17
Fix utc date time for end of the year cut off

### LINKED ISSUE #371: TIMEPERIODSUMMARY omitted dataset_count
This product: https://explorer.dev.dea.ga.gov.au/products/ls5_fc_albers is showing total of `33,482` datasets, in the explorer `time_overview` table adding all the `month` up total count is `33540`. It also lists `58` dataset_count against `start_day` `2011` but all the `center_time` are only from `2010` as per snippets below. The listed `2011` is also not showing up in the menu bar

image

```
odc=> select start_day, dataset_count, period_type from cubedash.time_overview where product_ref = (select id from cubedash.product where name = 'ls5_fc_albers') and dataset_count > 0;
 start_day  | dataset_count | period_type 
------------+---------------+-------------
 2010-11-01 |          3796 | month
 2010-12-01 |          4172 | month
 2010-01-01 |         33482 | year
 2011-01-01 |            58 | year
 2010-01-01 |          3776 | month
 2010-02-01 |          4332 | month
 2010-03-01 |          2120 | month
 2010-04-01 |          3804 | month
 2010-05-01 |          3071 | month
 2010-06-01 |          1430 | month
 2010-07-01 |           740 | month
 2010-08-01 |          1102 | month
 2010-09-01 |          2001 | month
 2010-10-01 |          3138 | month
 2011-01-01 |            58 | month
 1900-01-01 |         33482 | all
(16 rows)

odc=> select sum(dataset_count) from cubedash.time_overview where product_ref = (select id from cubedash.product where name = 'ls5_fc_albers') and dataset_count > 0 and period_type = 'month';
  sum  
-------
 33540
(1 row)


odc=> select count(*) from agdc.dataset where dataset_type_ref = (select id from agdc.dataset_type where name = 'ls5_fc_albers') and archived is NULL;
 count 
-------
 33540
(1 row)

odc=> select count(*) from agdc.dataset where dataset_type_ref = (select id from agdc.dataset_type where name = 'ls5_fc_albers') and archived is NULL and metadata->'extent'->>'center_dt' LIKE '2011%';
 count 
-------
     0
(1 row)


odc=> select count(*) from agdc.dataset where dataset_type_ref = (select id from agdc.dataset_type where name = 'ls5_fc_albers') and archived is NULL and metadata->'extent'->>'center_dt' LIKE '2010%';
 count 
-------
 33540
(1 row)
```

pindge:
fixes #371 

- add test case with selected 5 `ls5_fc_albers` datasets which exposes the tst issue
- fix tst issue
- revert accidental `docker-compose` port changes
- add `pytz` as a python library


jeremyh (approved):
Thanks for looking into this Pin! 

Ideally we keep using the `self.grouping_time_zone` on SummaryStore so that we have one consistent variable for time zone (rather than a new separate global), but given that things are already inconsistent this isn’t making it worse. Can do it in a future PR if desired.

pindge:
> Thanks for looking into this Pin!
> 
> Ideally we keep using the `self.grouping_time_zone` on SummaryStore so that we have one consistent variable for time zone (rather than a new separate global), but given that things are already inconsistent this isn’t making it worse. Can do it in a future PR if desired.

Yes, i have raised an issue to refactor  this, see issue #381

PR #382, opened 2022-06-01
align datasets for high tide comp centre_dt 1995 to 2000

### LINKED ISSUE #376: incorrect integration test case for high_low_tide
There are 76 metadata occurrences of 
```
extent:
    center_dt: '1995-01-01T00:00:00'
```
and 230 metadata occurrences of 
```
extent:
    center_dt: '2000-01-01T00:00:00'
```
in https://github.com/opendatacube/datacube-explorer/blob/develop/integration_tests/data/high_tide_comp_20p.yaml.gz

https://github.com/opendatacube/datacube-explorer/blob/develop/integration_tests/test_page_loads.py#L390-L392 

`306` is incorrect for the summary year `2008`, it should be `230`, and `76` for the year `2005`

pindge:
fixes #376


jeremyh:
This test data was dumped directly from our own (production) products. Have the original products changed?

robbibt (approved):
Looks good, thanks Pin!

PR #385, opened 2022-06-03
add test case for handling out of range datetime

### LINKED ISSUE #352: Some pages don't use the new datetime logic
(As reported by @alexgleith on Slack)

Some of Explorer's html templates call datacube's `Dataset.center_time` function directly to display the day. 

They should be updated to use the  [new day logic](https://github.com/opendatacube/datacube-explorer/pull/341) of Explorer's backend.

---

Can be reproduced by adding the dataset from Africa:
```
❯ datacube product add 'https://explorer.dev.digitalearth.africa/products/rainfall_chirps_daily.odc-product.yaml'
❯ datacube dataset add 'https://explorer.dev.digitalearth.africa/dataset/35cbccee-cb07-51cf-85d2-6d2948957544.odc-metadata.yaml'
```

Which has datestrings:
```
datetime: 2019-05-15T00:00:00Z
dtr:end_datetime: 2019-05-31T23:59:59Z
dtr:start_datetime: 2019-05-31T00:00:00Z 
```
(Note that the `datetime` doesn't fall inside the `start`/`end` time.. I don't know if that's valid, but it doesn't seem to be causing this issue.)

It incorrectly shows the 23rd of May on the dataset page as the date:

Screenshot from 2021-12-20 16-47-24

... but the time is correct in Explorer's database backend:

```
select * from cubedash.dataset_spatial where id ='35cbccee-cb07-51cf-85d2-6d2948957544';`
```

### LINKED ISSUE #383: need to add wrong center_dt datasets for integration test
PR #382 fixed #376, there need to be a specific test case to check explorer's behaviour with wrong `center_dt`

1. add a new product to integration test
2. add a few datasets with some containing wrong `center_dt`
3. test product summary

pindge:
Fixes: #352 and #383

Scope:
- [x] add template filter `center_time` to be fetched from dataset metadata
- [x] move `expects_eo3_metadata_type` function to `_utils.py`

Tests:
- [x] add a new product and 2 datasets testing `center_time` display
- [x] add test case to check displayed `center_time` is correct value from `cubedash.dataset_spatial` table
- [x] add test for region page `center_time`


cubedash/_pages.py
```diff
@@ -197,6 +197,11 @@ def search_page(
         _model.STORE.index.datasets.search(**query, limit=_HARD_SEARCH_LIMIT + 1),
         key=lambda d: d.center_time,
     )
+
+    for i in range(len(datasets)):
+        if _model.STORE.get_dataset_center_time(datasets[i].id):
+            datasets[i].center_time = _model.STORE.get_dataset_center_time(datasets[i].id)
```
jeremyh: This is going to result in hundreds of database queries, run once a time, before the page can be loaded. This is very slow on some environments, such as AWS+RDS where query latency is non-trivial.

(We faced a similar problem previously on the storage page on AWS, which initially had a query per product, which is far fewer queries than these two.)

If we do this, we should combine this loop into one single query, or return the info from the existing search query.

cubedash/_pages.py
```diff
@@ -303,6 +308,10 @@ def region_page(
         )
     )
 
+    for i in range(len(datasets)):
+        if _model.STORE.get_dataset_center_time(datasets[i].id):
+            datasets[i].center_time = _model.STORE.get_dataset_center_time(datasets[i].id)
+
```
jeremyh: (As above — too many queries)

cubedash/_pages.py
jeremyh: I’m also a little iffy about overwriting ODC’s own inner data models here directly. The datasets will in some areas of codebase have an Explorer center-time and others an ODC center-time, which might be surprising and unpredictable to future programmers.

In other places, we're keeping Explorer models separate from ODC models.

integration_tests/test_center_datetime_logic.py
```diff
@@ -0,0 +1,87 @@
...
+            assert created.type.name == "rainfall_chirps_daily"
+            dataset_count += 1
+        except AttributeError as ae:
+            assert dataset_count == 2
+            print(ae)
+    assert dataset_count == 2
+    return module_dea_index
+
+
+def test_datestring_on_dataset_page(client: FlaskClient):
```
jeremyh, line 48: :+1: love tests!

jeremyh (commented):
Thanks Pin.

A lot of this complexity is because we're trying to read it from the DB... but maybe we could calculate the same center time in Python on the front-end?

The [DB expression](https://github.com/opendatacube/datacube-explorer/blob/4afa0dbbb51d541f377c479e7edb914bdb62aef9/cubedash/summary/_extents.py#L481-L505) looks for a plain datetime before falling back to ODC's calculated center_time.
We could do the same thing in a flask filter when we display the center time. Then there's no need for modifying the original dataset or making extra queries.

jeremyh (approved):
:+1: Great! Thanks Pin.

Issue #387, opened 2022-06-05
region page should disable year summary if no dataset for region for the year


pindge:
i.e. https://explorer.digitalearth.africa/product/ls5_sr/regions/205050/1992

should be disabled 
![image](https://user-images.githubusercontent.com/24644475/172075674-b4b4d456-152e-41af-a01d-6239ebb55af4.png)

PR #388, opened 2022-06-07
extend region timeline overview

### LINKED ISSUE #305: Add timeline plot to region view in Explorer
Currently Explorer provides a very useful timeline feature at the top of Explorer to look at the amount of data that is available through time for an entire product:

image

It would be very useful if a similar plot could also be added to the region view that appears when a user clicks on a specific tile on the map (e.g. https://explorer.digitalearth.africa/product/ls5_sr/regions/205050). This would allow user to use Explorer to evaluate how much data is available through time for a specific location, rather than the entire product. 
This could look something like this:

image

pindge:
fix #305 and #387 

Scope:

- [x] Add a test product and datasets with regions
- [x] add assertion test to check product selection menu date displaying only date with datasets
- [x] add product region date selector filter for the selected region 
- [ ] add region summary years count for non-annual product


pindge:
needing some help from sqlalchemy https://github.com/sqlalchemy/sqlalchemy/discussions/8131

PR #390, opened 2022-06-08
make chart legend display diagonal

### LINKED ISSUE #389: timeline chart bar legend is cutting off the screen
https://explorer.dev.dea.ga.gov.au/products/ga_ls5t_ard_3/2005

image

with tooltip hovered

image
image

a better display is 

image

with tooltip hovered

image

image

adding `rotate(45deg)` to css

```
    z-index: 1;
    position: absolute;
    bottom: 0;
    text-align: left;
    font-size: 10px;
    font-weight: normal;
    color: #333;
    transform: translateY(calc(100% + 5px)) rotate(45deg);
    transform-origin: top left;
```

pindge:
fix #389


robbibt (approved):
Hey Pin, I think this looks OK - although we may need to see what it looks like once this is merged then tweak things if it looks weird!

Saw some tests fail below, not sure if that's a problem

PR #394, opened 2022-06-15
make explorer to record region_code in metadata as first preference

### LINKED ISSUE #342: Incorrect tile numbers displayed for DEA C3 summary products
The DEA C3 summary product grid,  [defined here](https://docs.dea.ga.gov.au/reference/collection_3_summary_grid.html) is not being correctly represented within the explorer map UI [example here](https://explorer.dev.dea.ga.gov.au/products/ga_ls_wo_fq_nov_mar_3/extents)
Clicking on the tile in the top left (Tile -17, -11) you are actually accessing tile x11y46 [shown here](https://explorer.dev.dea.ga.gov.au/product/ga_ls_wo_fq_nov_mar_3/regions/-17_-11)

This tile index is being infered from the product storage information, which be believe to be correct [here](https://explorer.dev.dea.ga.gov.au/products/ga_ls_wo_fq_nov_mar_3) not from the dataset region code (or equivalent) [visible here](https://explorer.dev.dea.ga.gov.au/products/ga_ls_wo_fq_nov_mar_3/datasets/94594ba3-ecf6-4894-9825-5b985d260dda)

pindge:
fix #342 
if metadata provides region_code, that should be used directly, only if it's not offered, do the additional calculation.

- [x] add test case as raised by issue #342, a case where metadata provides region_code and it displays correctly
- [x] add test case where the region is defined in the metadata but all are the same,  explorer should still display the region as provided by the  metadata, this will help expose issues during the development phase 
- [x] add test case where the region is not defined in the metadata and explorer needs to calculate the tiles


jeremyh (approved):
Thanks Pin. This sounds like a good idea now that ingested products are mostly defunct.

We'll need to mention on release that this change can make affected products inconsistent (summaries could exist with both old regions and new regions depending on when they were generated). 

People will need to `--force` their summary update for those products to recreate all summaries consistently with the new regions.

pindge:
> Thanks Pin. This sounds like a good idea now that ingested products are mostly defunct.
> 
> We'll need to mention on release that this change can make affected products inconsistent (summaries could exist with both old regions and new regions depending on when they were generated).
> 
> People will need to `--force` their summary update for those products to recreate all summaries consistently with the new regions.

noted!

PR #396, opened 2022-06-17
remove region from region table if all relating datasets are archived

### LINKED ISSUE #395: Region pages are still accessible after all the datasets relating to the region are archived
The region page should no longer be accessible after all the datasets relating to that region are archived. should display 404

this SQL logic does not handle the case all the datasets for the region are archived or deleted, resulting it staying in the database
https://github.com/opendatacube/datacube-explorer/blob/a26309d03a6b8830c31d59f3bba2bcad4b8f43fb/cubedash/summary/_stores.py#L545-L574

pindge:
Fixes #395 

## Scope

- extend existing test case to include  testing for handling of region after all datasets are archived for the region
- extend region SQL to delete the row if no matching region is in `dataset_spatial`

### DIFF
--- a/cubedash/summary/_stores.py
+++ b/cubedash/summary/_stores.py
@@ -533,9 +533,7 @@ class SummaryStore:
             last_refresh_time=covers_up_to,
         )
 
-        # TODO: This is an expensive operation.
-        # We regenerate them all every time there are changes.
-        # pins' bookmark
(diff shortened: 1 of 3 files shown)


jeremyh (approved):
Looks good. Love that you’re always expanding tests. 

I am curious if this will slow down the summary-gen much. There’s room to optimise if so. We’ll find out. Thanks Pin.

PR #398, opened 2022-06-20
fix find-years-needing-update logic where product count is none but residue years still recorded

### LINKED ISSUE #397: archived datasets displaying for the years in dataset_count audit page
image

archived datasets are still being listed on the csv available from dataset count audit page. i.e. https://explorer-aws.dea.ga.gov.au/audit/dataset-counts

The menu link was still showing for 2020, but going on to the day, there are zero datasets, which means the timeperiodoverview wasn't cleaned out properly

This was corrected after running `--force-refresh`

pindge:
fixes #397 

## Scope
- add test containing 1 dataset to test archival residues
- add `id` to HTML for dataset-count audit page for easy HTML test
- add integration test to check dataset count on audit page for both before and after archival is performed
- fix `find_years_needing_update` function where `product.dataset_count` is `0`, but there used to be `years` recorded with datasets.

### DIFF
--- a/cubedash/summary/_stores.py
+++ b/cubedash/summary/_stores.py
@@ -374,22 +374,12 @@ class SummaryStore:
 
         Either:
            1) They don't exist yet, or
-           2) They have month-records that are newer than our year-record.
+           2) They existed before and has been deleted or archived, or
+           3) They have month-records that are newer than our year-record.
         """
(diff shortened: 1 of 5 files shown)


jeremyh (approved):
Love it. Thanks Pin

PR #401, opened 2022-07-07
[UI] hide duplicate field creation_dt from dataset page

### LINKED ISSUE #386: dataset page displays duplicated fields
on `/products/<product_name>/datasets/<uuid>` page, in the `Fields` section displays `creation_dt` and `creation_time`. This appears to be duplicated information

image

pindge:
Fixes #386 

`creation_time` was added to datacube-core as per https://github.com/opendatacube/datacube-core/pull/432, explorer shouldn't display duplicate fields on the dataset page.


cubedash/templates/layout/macros.html
```diff
@@ -52,7 +52,7 @@
 {% macro query_param_list(d, prefix="", wide=false, show_nulls=true, show_dicts=true, descriptions=None, fallback_dict={}) -%}
     <div class="query-param-list {% if wide %}wide{% endif %}">
         {% for key, value in d | dictsort if (show_dicts or value is not mapping) recursive %}
-            {% if show_nulls or (value is not none) or (key in fallback_dict) %}
+            {% if (show_nulls or (value is not none) or (key in fallback_dict)) and (key != "creation_dt") %}
```
jeremyh, line 56: Not essential, but I do like to leave comments on lines like this, as it will be a head-scratcher for future programmers who see it.

```suggestion
            {# "creation_dt" is excluded, as it's a duplicate of the creation_time field (latter is newer and date typed) #}
            {% if (show_nulls or (value is not none) or (key in fallback_dict)) and (key != "creation_dt") %}
```

jeremyh (approved):
Thanks Pin!

PR #407, opened 2022-07-11
Display no matching region product with grey color in product selection 

### LINKED ISSUE #57: User can navigate to regions that don't exist
When viewing a specific region, eg: https://data.dea.gadevs.ga/region/ls8_nbar_albers/20_-34

The product drop down menu lets you switch to any other product, even though most products don't have that region defined.

The resulting page currently throws an exception, such as with [CUBEDASH-1G](https://sentry.io/organizations/jeremy-hooke/issues/924840570/?project=1272110&referrer=slack&statsPeriod=14d)

We should:
- make it clear to the user which products they can navigate to for the current region: either filter the product list or grey-out those without the region
- Return an error message (404 / 400?) instead of an exception when loading missing regions.

pindge:
Fixes #57 

## Scope: 
- add a test case for checking the product selection menu while on the region page, only products containing the same region will display in white and every other product display in grey (indicating empty)
- add product `ls7_nbart_tmad_annual` and one dataset sharing the same region_code as a dataset in `ls5_nbart_tmad_annual`
- if on region pages, product selection will grey out all products without matching region_code
- On any other page, only products with empty datasets will be greyed out.

image

### DIFF
--- a/cubedash/summary/_extents.py
+++ b/cubedash/summary/_extents.py
@@ -622,6 +622,37 @@ def datasets_by_region(
     )
 
 
+def products_by_region(
+    engine: Engine,
+    index: Index,
+    region_code: str,
+    time_range: Range,
+    limit: int,
+    offset: int = 0,
+) -> Generator[Dataset, None, None]:
+    query = (
+        select(postgres_api._DATASET_SELECT_FIELDS)
+        .select_from(
+            DATASET_SPATIAL.join(DATASET, DATASET_SPATIAL.c.id == DATASET.c.id)
+        )
+        .where(DATASET_SPATIAL.c.region_code == bindparam("region_code", region_code))
+    )
+    if time_range:
+        query = query.where(
+            DATASET_SPATIAL.c.center_time > bindparam("from_time", time_range.begin)
+        ).where(DATASET_SPATIAL.c.center_time < bindparam("to_time", time_range.end))
+    query = (
+        query.order_by(DATASET_SPATIAL.c.center_time)
+        .limit(bindparam("limit", limit))
+        .offset(bindparam("offset", offset))
+    )
+
+    return (
+        res.dataset_type_ref
+        for res in engine.execute(query).fetchall()
+    )
+
+
 @dataclass
 class RegionSummary:
     product_name: str
--- a/cubedash/summary/_extents.py
+++ b/cubedash/summary/_extents.py
@@ -638,7 +638,7 @@ def products_by_region(
         query = query.where(
             DATASET_SPATIAL.c.center_time > bindparam("from_time", time_range.begin)
         ).where(DATASET_SPATIAL.c.center_time < bindparam("to_time", time_range.end))
-    
+
     query = (
         query.order_by(DATASET_SPATIAL.c.dataset_type_ref)
         .limit(bindparam("limit", limit))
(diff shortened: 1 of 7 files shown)


cubedash/summary/_extents.py
```diff
@@ -622,6 +622,37 @@ def datasets_by_region(
...
+    query = (
+        query.order_by(DATASET_SPATIAL.c.center_time)
+        .limit(bindparam("limit", limit))
+        .offset(bindparam("offset", offset))
+    )
+
+    return (
+        res.dataset_type_ref
+        for res in engine.execute(query).fetchall()
+    )
```
jeremyh: This query will be _very_ slow on our main instances. And it returns one-row-per-dataset rather than one-per-product, which might be a mistake? (the name says "find product". And when it hits the row limit from datasets, will be missing products in its return value)

Example improvement:

```suggestion
) -> Generator[DatasetType, None, None]:
    query = (
        select([DATASET_SPATIAL.c.dataset_type_ref]).distinct()
        .where(DATASET_SPATIAL.c.region_code == bindparam("region_code", region_code))
    )
    if time_range:
        query = query.where(
            DATASET_SPATIAL.c.center_time > bindparam("from_time", time_range.begin)
        ).where(DATASET_SPATIAL.c.center_time < bindparam("to_time", time_range.end))
    
    query = (
        query.order_by(DATASET_SPATIAL.c.dataset_type_ref)
        .limit(bindparam("limit", limit))
        .offset(bindparam("offset", offset))
    )

    return (
        index.products.get_by_name(res.dataset_type_ref)
        for res in engine.execute(query).fetchall()
    )
```

Ie.
- Return distinct products, rather than the product repeated for every dataset. 
- Remove the join (since we're throwing away all dataset information anyway)
- Only select the `dataset_type_id` (so we don't need to transfer so much data from the db -- we're throwing the rest away)
- Return a DatasetType object (ie. ODC's Product), since it's called `find_product()`.

... although we'll need to update the other code to expect `DatasetType.id` rather than plain id.

---

... but this query still may not be fast enough on our bigger instances? We could add a dedicated index for it, possibly:

Eg.
```
$ select distinct dataset_type_ref from cubedash.dataset_spatial where region_code='53JMN' limit 100;
 dataset_type_ref 
══════════════════
               90
               91
(2 rows)

Time: 212.347 ms
$ create index region_code_products on cubedash.dataset_spatial(region_code, dataset_type_ref);
$ select distinct dataset_type_ref from cubedash.dataset_spatial where region_code='53JMN' limit 100;
 dataset_type_ref 
══════════════════
               90
               91
(2 rows)

Time: 16.291 ms
```
(was after warmup)

cubedash/summary/_stores.py
```diff
@@ -1643,6 +1643,28 @@ def find_datasets_for_region(
...
 
+    def find_products_for_region(
+        self,
+        region_code: str,
+        year: int,
+        month: int,
+        day: int,
+        limit: int,
+        offset: int = 0,
+    ) -> Iterable[Dataset]:
```
jeremyh: Doesn't return `Iterable[Dataset]` at the moment (it's returning strings)... maybe `Iterable[DatasetType]` if we intend it to "find products"? (see other comment)

jeremyh (commented):
Thanks Pin, nice idea!

jeremyh (approved):
Thanks Pin. We'll keep an eye on performance after deployment. If its too slow we should add an index.

PR #420, opened 2022-08-03
add localised time and utc time on tooltip

### LINKED ISSUE #419: dataset search page time not localised
when dataset timezone is summarised by explorer its summarised with local timezone, the search page still displays dataset datetime in UTC as per dataset metadata

https://explorer.dev.dea.ga.gov.au/products/ls5_fc_albers/datasets/2011/1

image

pindge:
fixes #419, #420 

## Scope
- add tooltip for displaying time in UTC equiv
- make search page for datasets display in local time
- add flask app config for `CUBEDASH_DEFAULT_TIMEZONE`
- add `CUBEDASH_DEFAULT_TIMEZONE` config to README
- make `center_time_from_metadata` filter to always return time in `default_utc`
- Change column name `Time` to `Time (local)` for clarity
- add test case for checking both localised time and UTC time on tooltip 

## build fix
latest flask library is broken, pin to `2.1.3`

## Visual change

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,8 @@ You can alter default [Flask](http://flask.pocoo.org/docs/1.0/config/) or
     # Maximum number of source/derived datasets to show
     CUBEDASH_PROVENANCE_DISPLAY_LIMIT = 20
 
+    CUBEDASH_DEFAULT_TIMEZONE = "Australia/Darwin"
+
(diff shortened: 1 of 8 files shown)


robbibt (approved):
This looks amazing @pindge, very exciting!

PR #422, opened 2022-08-05
Default timezone

### LINKED ISSUE #381: Make astimezone configuration application configurable
Proposed solution: 
1. add extra logic to check if `total dataset count` matches the `summary total count`, if not match add more years to include the rolled over
2. make `Australia/Darwin` configurable for `africa` and `antarctica` projects

_Originally posted by @pindge in https://github.com/opendatacube/datacube-explorer/issues/371#issuecomment-1127130206_

## Details

Current explorer application has hardcoded the `grouping_time_zone` to `Australia/Darwin`, this needs to be application configurable https://github.com/opendatacube/datacube-explorer/blob/9d6222a3695d98514ed6244daee0d3301ddca1b1/cubedash/summary/_summarise.py#L45

pindge:
fixes #381 

## Scope
- add test case for generate cli to run for configured timezone summary generation
- add test case for testing configured timezone summary search results
- remove duplicate hard coding of `Australia/Darwin` and set all others to import from default
- all flask page are set to use `app.config["CUBEDASH_DEFAULT_TIMEZONE"]`

----
:books: Documentation preview :books:: https://datacube-explorer--422.org.readthedocs.build/en/422/


JonDHo:
Hi. Any update on this PR? This causes lots of issues for implementations in timezones which are very different from Austaralia/Darwin (e.g. all of the Americas) because the hard-coded grouping timezone ends up shifting the dates of capture into the following day, creating a lot of confusion for local users.

whatnick:
I will be interested in taking over this PR to progress slowly. The first point of call would be to resolve conflicts and refresh it.

JonDHo:
I did a local merge and there were only a couple of conflicts which were pretty simple to resolve. I have been testing it today but haven’t managed to get it to do what I was hoping. Might be worth a direct catch-up to show you the issues at some point if you want.

whatnick:
Should address pre-commit gripes around Python Flake8 behaviour in a separate PR.

JonDHo:
One issue I have found is that when daylight savings changes occur during a month, the Summariser process can fail when the time step does not exist. This happens because when the date range is generated at [L137 in _summarise.py](https://github.com/opendatacube/datacube-explorer/blob/998d6a52880d5757cedd099c2588d67a4bad949a/cubedash/summary/_summarise.py#L136-L138), the range can end up including a time which doesn't exist. When the timezone changes, midnight on the date of change doesn't actually exist because it is instantly 1am. To resolve this, pd.date_range needs to include `nonexistent="shift_forward"` or other suitable setting, which is ultimately passed on to the underlying pandas function (tz_localize_utc). I can't actually find tz_localize_utc, but assume it is similar to https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.dt.tz_localize.html

Code to repeat error (derived/modified from /cubedash/summary/_summarise.py):
```python
import pandas as pd
from dateutil import tz
import datetime
def _with_default_tz(d: datetime) -> datetime:
  if d.tzinfo is None:
    return d.replace(tzinfo=_grouping_time_zone_tz)
  return d

_grouping_time_zone_tz = tz.gettz("America/Santiago")

begin_time=_with_default_tz(datetime.datetime(2016, 8, 1, 0, 0))
end_time=_with_default_tz(datetime.datetime(2016, 9, 1, 0, 0))

pd.date_range(begin_time, end_time, inclusive="left")
```
Replacing the final line with the following line fixes the issue:
```python
pd.date_range(begin_time, end_time, inclusive="left", nonexistent="shift_forward")
```

SpacemanPaul (commented):
I'm not that familiar with the explorer code base.  This PR looks like a reasonable incremental change to me, and @JonDoHo's suggested tweak makes sense.

But IMO it shouldn't be necessary to deal with statutory timezones at all.  There should only two grouping rules needed: 

1) solar_day grouping, which is baked into ODC search already.
2) UTC calendar-day grouping (which is trivial to implement.)

This distinction would not be an issue in the Americas - but must be an issue in Australia.  How does Explorer know to not use time-zone based grouping for summary products with midnight UTC dates?  Is there a list of summary products in config somewhere or something?

JonDHo:
The issue isn’t the data itself, but the representation of the times in a user’s local timezone, which is based on the user’s location set by their browser. What happens is that explorer by default uses midnight as the timestamp for day steps (actually time is missing, which is interpreted as 0, or midnight on the specified day). Of course the real times do exist, but they are not used for much (all?) of the summaries. The month summaries are simply based on a start & end day for the month from which a range is generated. 

In chile, or any timezone behind UTC, subtracting hours (e.g. UTC-3) from midnight results in data incorrectly showing as the day before. In reality, data is collected at around 14:00 UTC (for Sentinel 2), so the correct time should be about 11:00 chile time, not 21:00 the day before.

In short though, time handling sucks.

JonDHo:
Oh, I should add that when summarising the data, to have the year/month summaries match the day summaries, it is necessary to define a timezone. Day summaries are calculated on page load, in the user’s timezone, but year and month summaries are generated by cubedash-gen routinely. If you don’t set a timezone for year/month, it defaults to Australia/Darwin timezone. There can obviously be problems with users in different time zones, but no matter what, it shouldn’t default to Darwin. Generating summaries in UTC might be the simplest option, but for a national system with a single timezone (eg. Chile), using a local tz might be sensible.

SpacemanPaul (approved):
Still not convinced we're using the right approach here, but realistically we are stuck with it for the time being.  Given that, approved.

JonDHo:
Yep. Completely agree. I think that solar day is actually the right approach but this at least gives us all flexibility for now.

whatnick:
Unfortunately this cannot be merged due to conflicts. I will try to rebase to develop and resolve. Help here is appreciated. @JonDHo if you could pull it into a fork and raise a fresh PR that would be nice. All changes should be present in this branch.

I have spent sometime rebasing all the changes. Seems messy especially given conflicting changes in `conftest.py` and `test_utc_tst.py`.

Process to rebase to develop to follow for commits :

```sh
git fetch
git pull
git rebase develop
# Complains about conflicts for each commit, resolve those sanely
git rebase --continue
# Go over all commits on the branch until git is happy
git commit
git push --force-with-lease
```

[Rebasing](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) is tricky as is, I cannot unlock the repo to allow Squash merge which would be the easy way out.

whatnick:
This has been squash merged and repository reverted to rebase merge. In the future PR's will need to be kept refreshed and shepherded with regular rebase to `develop` to avoid the need for extreme measures. We can recover from here as appropriate.

Issue #431, opened 2022-08-24
Incorrect summary displayed on map page for some fields (at least maturity)


omad:
Within DEA we have just switched to generating `interim` quality ARD data instead of `final` quality for Landsat 8, due to the recent unavailability of BRDF.

When I go looking for this information, Explorer is presenting an inconsistent view.

On the dataset listing page, I correctly see a list of both `final` and `interim` datasets:
![image](https://user-images.githubusercontent.com/108979/186341279-5ae79207-1beb-424e-8275-fd9caf9deb77.png)

However, when I view the same date on the Map/Summary page, it indicates that all datasets are of `final` quality.
![image](https://user-images.githubusercontent.com/108979/186341331-c18dc0e6-7689-43c6-9881-a83e200dcede.png)

jeremyh:
Yeah, it is a little misleading. The little squiggly underline on the word is because Explorer is guessing -- it's based on a sampling of the datasets. So all datasets it sampled consistently had that property.

There's a little tooltip:
![Screenshot from 2022-08-25 09-49-40](https://user-images.githubusercontent.com/25688/186543750-e8e5c3f0-1529-4098-92f8-5d5b5dbaaf79.png)

The non-squiggly-underlined ones are actually concrete rules on the product

pindge:
## Proposed solution
1. add a description to the block, indicating the sample size
2. instead of comparing with the first dataset, search for the latest indexed datasets https://github.com/opendatacube/datacube-explorer/blob/a00bcb04884d4350f2a21f9a4cfa96ed8d008c09/cubedash/summary/_stores.py#L631

PR #432, opened 2022-08-31
setup test case for dataset common fields

### LINKED ISSUE #433: Incorrect logic for sample datasets used for fixed_metadata
current sampling logic: 
https://github.com/opendatacube/datacube-explorer/blob/develop/cubedash/summary/_stores.py#L651-L654

running the sql based on the above logic returns an inconsistent sample size.
```
opendatacube_test=# select count(*) from agdc.dataset tablesample system (100);
 count 
-------
     7
(1 row)

opendatacube_test=# select count(*) from agdc.dataset tablesample system (50);
 count 
-------
     7
(1 row)

opendatacube_test=# select count(*) from agdc.dataset tablesample system (40);
 count 
-------
     0
(1 row)
```

this is further validated by test ci: https://github.com/opendatacube/datacube-explorer/runs/8104066958?check_suite_focus=true,
sample percentage of `50` returned zero row of datasets

```
def test_product_fixed_metadata_by_sample_percentage(summary_store: SummaryStore):
        # There are 4 interim and 16 final maturity level datasets
        # at [100](https://github.com/opendatacube/datacube-explorer/runs/8104095953?check_suite_focus=true#step:5:101)% (all 20 datasets), the same dictionary will be returned
        # 100% of the time
        fixed_fields = summary_store._find_product_fixed_metadata(
            summary_store.index.products.get_by_name("ga_ls8c_ard_3"),
            sample_percentage=100,
        )
    
        assert fixed_fields == {
            "platform": "landsat-8",
            "instrument": "OLI_TIRS",
            "product_family": "ard",
            "format": "GeoTIFF",
            "eo_gsd": 15.0
        }
    
        # There are 4 interim and 16 final maturity level datasets
        # at 50% (10 datasets), there is a fair chance, that maturity level
        # will be in the dictionary
        fixed_fields = summary_store._find_product_fixed_metadata(
            summary_store.index.products.get_by_name("ga_ls8c_ard_3"),
            sample_percentage=50,
        )
    
>       assert len(fixed_fields) >= 5
E       assert 0 >= 5
E        +  where 0 = len({})
```

pindge:
Fixes: #433 and #431 

## Scope
- rework logic for datasets sampling as `tablesample` logic can return zero rows and its much more expensive to run compared with `order by random`
- add test case for different sampling percentage 

## UI Changes
- add heading for common fields and hover over information
image
image

## Performance Comparison

```
odc=> select count(*) from agdc.dataset tablesample system (0.2) where dataset_type_ref = 1;
 count 
-------
    46
(1 row)

odc=> EXPLAIN select count(*) from agdc.dataset tablesample system (0.2) where dataset_type_ref = 1;
                             QUERY PLAN                              
---------------------------------------------------------------------
 Aggregate  (cost=29393.86..29393.87 rows=1 width=8)
   ->  Sample Scan on dataset  (cost=0.00..29393.81 rows=19 width=0)
         Sampling: system ('0.2'::real)
         Filter: (dataset_type_ref = 1)
(4 rows)

odc=> \q^C
odc=> EXPLAIN select id from agdc.dataset where dataset_type_ref  =1 order by random() limit 46;
                                                     QUERY PLAN                                                     
--------------------------------------------------------------------------------------------------------------------
 Limit  (cost=24471.98..24472.09 rows=46 width=24)
   ->  Sort  (cost=24471.98..24495.13 rows=9260 width=24)
         Sort Key: (random())
         ->  Index Scan using ix_agdc_dataset_dataset_type_ref on dataset  (cost=0.43..24169.94 rows=9260 width=24)
               Index Cond: (dataset_type_ref = 1)
(5 rows)
```


jeremyh (approved):
Looks good, thanks Pin!

We could also squash those two queries back into one query to keep latency smaller, but this is a good improvement as-is.

PR #436, opened 2022-09-02
render dataset count table as data table

pindge:
## Scope

- render dataset count page table with datatable jquery plugin

*on load*
image

*order by highest ds count*
image

*search by matching char*
image


robbibt (approved):
So cool, looks great to me! Another thing that could be handy in the future would be the ability to sort by "Last updated" or similar, to let users sort products by old/new

jmettes (approved):

jmettes:
Very handy!

PR #447, opened 2022-10-17
Sentry sdk

### LINKED ISSUE #441: Raven is deprecated in favor of Sentry-Python.
as per raven-python repo (https://github.com/getsentry/raven-python), this library is deprecated
this  library will enable `env` tagging

image

pindge:
Fixes #441 

- update doc for sentry setup
- remove raven.js test case
- remove raven and install sentry-sdk

### DIFF
--- a/cubedash/templates/500.html
+++ b/cubedash/templates/500.html
@@ -15,13 +15,4 @@
     This has been logged.
 </div>
 </body>
-{% if sentry_event_id %}
-    <script src="https://cdn.ravenjs.com/3.26.4/raven.min.js" crossorigin="anonymous"></script>
-    <script>
-        Raven.showReportDialog({
-            eventId: '{{ sentry_event_id }}',
-            dsn: '{{ sentry_public_dsn }}'
-        });
-    </script>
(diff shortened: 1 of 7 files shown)


cubedash/templates/500.html
```diff
@@ -15,13 +15,4 @@ <h4>Internal error</h4>
     This has been logged.
 </div>
 </body>
-{% if sentry_event_id %}
```
whatnick, line 18: Should add new JS snippet here : https://docs.sentry.io/platforms/javascript/install/cdn/ . Creating separate front end fix issue for it.

integration_tests/test_page_loads.py
```diff
@@ -72,17 +72,6 @@ def _script(html: HTML):
     return html.find("script")
 
 
-def test_sentry(sentry_client: FlaskClient):
```
whatnick, line 75: This test will need to be updated for front end change.

whatnick (approved):
Let's do the remove and add back. Since this is optional functionality.

PR #455, opened 2022-10-18
simplify docker build

pindge:
- remove extra step for `COPY` and `pip install`


Dockerfile
```diff
@@ -63,10 +57,10 @@ ADD . $APPDIR
 # then we want to link the source (with the -e flag) and if we're in prod, we
 # want to delete the stuff in the /code folder to keep it simple.
 RUN if [ "$ENVIRONMENT" = "deployment" ] ; then\
-        pip install .[$ENVIRONMENT] ; \
+        pip install .[$ENVIRONMENT] --extra-index-url https://packages.dea.ga.gov.au/; \
```
jmettes, line 60: Does it make sense for an ODC repo to depend on DEA infra?

pindge: Its being installed regardless, see `requirement-docker.txt`, this pr is simplify removing that `.txt` and put them together

jmettes (approved):

jeremyh:
The advantage of two-step install was that code changes don't need a full rebuild of all dependencies.

Docker can cache the first step, containing all built dependencies.

Then code changes only need to rerun the (very fast) second step to update your container during development.

PR #458, opened 2022-10-18
set github action to test prod image

### LINKED ISSUE #456: Explorer lacks test for deployment docker image
the #447 pr passed github ci tests but the deployment image is erroring due to ```Error: No module named 'sentry_sdk'```

There needs to be a ci test for deployment image

pindge:
Fixes: #456

### DIFF
--- a/.github/workflows/deployment_test.yaml
+++ b/.github/workflows/deployment_test.yaml
@@ -29,6 +29,11 @@ jobs:
           make init-odc
           make schema
 
+      - name: Sleep for 10 seconds
+        uses: whatnick/wait-action@master
+        with:
+          time: '5s'
+
       - name: Test ping
         run: |
           curl --show-error --fail \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -110,28 +110,28 @@ up-prod: ## Start using the prod Docker image
 		up -d
 
 init-odc: ## Initialise ODC Database
-	docker-compose exec -T explorer /bin/sh -c \
-		"datacube system init"
+	docker-compose run -T explorer \
```
alexgleith: `run` does something different to `exec`

You want to use `exec`, as it executes something in the running container. Run creates a new container.

.github/workflows/deployment_test.yaml
```diff
@@ -0,0 +1,49 @@
+---
+name: Deployment Docker Image Tests
```
alexgleith:
```suggestion
name: Docker Image
```

alexgleith (approved):
Looks good to me.

One suggestion about the name of the workflow.

And a suggestion for later maybe about caching the build, which will speed it up. 3 minutes isn't too bad though.

SpacemanPaul (approved):
LGTM

PR #487, opened 2022-11-17
Add dataset thumbnail

### LINKED ISSUE #475: Include dataset thumbnails on Explorer dataset pages
Our new automatically generated products now have beautiful thumbnail JPEGs for each dataset that show at a glance what the data looks like, e.g.
https://explorer-aws.dea.ga.gov.au/products/ga_ls8c_ard_3/datasets/558f9cef-91c7-402d-ac1d-10a4b69917bd
https://data.dea.ga.gov.au/baseline/ga_ls8c_ard_3/107/079/2013/09/01/ga_ls8c_nbart_3-0-0_107079_2013-09-01_final_thumbnail.jpg

It would be fantastic to expose these images in Explorer, so that users can easy inspect a dataset without having to download it onto their PC. For example, this could look something like this, with the thumbnail displayed next to the map that shows where the data is located in Australia:

image

Ariana-B:
Add image overlay to map if nbart thumbnail link is found.

Tested by running locally; seems to work as expected.

image

----
:books: Documentation preview :books:: https://datacube-explorer--487.org.readthedocs.build/en/487/

### DIFF
--- a/cubedash/templates/dataset.html
+++ b/cubedash/templates/dataset.html
@@ -204,7 +204,7 @@
                     ' &copy; <a href="https://cartodb.com/attributions">CartoDB</a>'
                 }
             );
-            
+            var coords = [];
             const dataset_data = L.geoJson({{ extent_geojson | torapidjson }},
                 {
                     style: function (feature) {
@@ -230,10 +230,7 @@
             map.fitBounds(dataset_data.getBounds(), {animate: false, maxZoom: 5});
             const paths = '{{ (dataset|dataset_file_paths)|tojson }}';
             if (JSON.parse(paths)['thumbnail:nbart']) {
-                dataset_data.bindPopup(
-                    `<img width="200" height="200" src="{{ thumbnail_url[0]|e }}">`,
-                    { offset: L.point(0, -20)}
-                ).openPopup();
+                L.imageOverlay("{{ thumbnail_url[0]|e }}", dataset_data.getBounds()).addTo(map);
             }
 
             window.MAP = map;
--- a/cubedash/templates/dataset.html
+++ b/cubedash/templates/dataset.html
@@ -144,13 +144,14 @@
                                     <i class="fa fa-caret-right" aria-hidden="true"></i>
                                 </th>
                                 <td>
-                                    <a href="{{ offset | resolve_remote_url(uri) }}"
+                                    {% set offset_url = offset | resolve_remote_url(uri) %}
+                                    <a href="{{ offset_url }}"
                                        class="uri-path">
                                         {{ offset }}
-                                        {% if 'thumbnail:nbart' == name %}
-                                            {# workaround for jinja preventin global varaibles from being updated inside
+                                        {% if 'thumbnail:nbart' == name or 'thumbnail' == name %}
+                                            {# workaround for jinja preventing global varaibles from being updated inside
                                                 for loops and if/else statements #}
-                                            {% set _ = thumbnail_url.append(offset | resolve_remote_url(uri)) %}
+                                            {% set _ = thumbnail_url.append(offset_url) %}
                                         {% endif %}
                                     </a>
                                 </td>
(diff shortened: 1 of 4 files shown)


robbibt:
This looks SO cool @Ariana-B, exactly what I had in mind!

One question/suggestion: I can see in the code that it appears to only plot the thumbnail if it's from an NBART product. Would it be possible to generalise this to plot thumbnails for any product if they're available? For example, we also have thumbnails for most of our other Collection 3 products, and it'd be fantastic to be able to visualise them too. 

e.g. all of these (and more) have a "_thumbnail.jpg":

https://data.dea.ga.gov.au/?prefix=derivative/ga_ls_fc_3/2-5-1/092/076/2018/06/17/
https://data.dea.ga.gov.au/?prefix=derivative/ga_ls_mangrove_cover_cyear_3/3-0-0/x27/y44/2001--P1Y/
https://data.dea.ga.gov.au/?prefix=derivative/ga_ls_wo_fq_apr_oct_3/1-6-0/x32/y28/1998-04--P7M/
https://data.dea.ga.gov.au/?prefix=derivative/ga_ls_tc_pc_cyear_3/1-0-0/x27/y36/1991--P1Y/

omad:
This is looking great Ariana. I was expecting it to require Python/Flask modifications as well as the template, hopefully the logic doesn't get too hairy by keeping it in the template with my next request.

The DEA ARD data include thumbnails under the `nbart:thumbnail` tag, but most datasets that include a thumbnail are simply `thumbnail`. Can you please update this to also show the thumbnail image for those types of datasets.

Eg. 
- [ga_ls_wo_fq_cyear_3](https://explorer.sandbox.dea.ga.gov.au/products/ga_ls_wo_fq_cyear_3/datasets/6d5327b2-f173-48f0-b719-734f86eae50c)
- [ga_ls_fc_3](https://explorer.sandbox.dea.ga.gov.au/products/ga_ls_fc_3/datasets/a9a8a82c-0a33-5d70-a56f-db1f604e3c99)
- [ga_ls_mangrove_cover_cyear_3](https://explorer.sandbox.dea.ga.gov.au/products/ga_ls_mangrove_cover_cyear_3/datasets/d89ac4d0-4769-4c34-ae10-537ff46a0de6)

Ariana-B:
@omad @robbibt 
updated to include `thumbnail` tag

cubedash/templates/dataset.html
```diff
@@ -139,9 +144,15 @@ <h3>
...
-                                    <a href="{{ offset | resolve_remote_url(uri) }}"
+                                    {% set offset_url = offset | resolve_remote_url(uri) %}
+                                    <a href="{{ offset_url }}"
                                        class="uri-path">
                                         {{ offset }}
+                                        {% if 'thumbnail:nbart' == name or 'thumbnail' == name %}
+                                            {# workaround for jinja preventing global varaibles from being updated inside
+                                                for loops and if/else statements #}
+                                            {% set _ = thumbnail_url.append(offset_url) %}
+                                        {% endif %}
```
omad: Eeep, workarounds :(

See below for a nicer option, but open to your thoughts to.

cubedash/templates/dataset.html
```diff
@@ -219,6 +229,10 @@ <h3 class="followed">
             });
             L.control.zoom({position: "bottomright"}).addTo(map);
             map.fitBounds(dataset_data.getBounds(), {animate: false, maxZoom: 5});
+            const has_thumbnail = '{{ thumbnail_url|length >= 1 }}';
```
omad: How about instead of the work around with a `thumbnail_url` list. We implement a new jinja2 filter for extracting a thumbnail URL from a dataset.

It will be ever so slightly less efficient, but IMO cleaner and less likely to get broken in the future if/when these templates get refactored.

See https://github.com/opendatacube/datacube-explorer/blob/develop/cubedash/_filters.py#L120-L122 for the implementation of the `dataset_file_paths` filter used above.

I'm ambivalent about keeping the implementation in `_utils.py` and the filter in `_filters.py`, I don't see why some filters can contain the own implementation... but on the other hand, it would keep the code more consistent to continue with that pattern. 🖐

omad (commented):
What do you think about moving the implementation out of the template and into a filter. See 👇 for details.

cubedash/templates/dataset.html
Ariana-B: Agreed; done

omad (approved):
Looks awesome, thank you!

robbibt:
Thanks @Ariana-B! Super looking forward to seeing this go live - can someone tag me when that happens so I can test it out?

PR #488, opened 2022-11-21
Make it easier to browse dataset s3 files

### LINKED ISSUE #480: Make it easier to browse files on S3 for a given dataset
A common use case is locating a certain dataset on Explorer, then wanting to do a deeper dive to inspect this dataset by looking into its files stored on S3. This can be particularly useful for accessing files that aren't exposed by Datacube (e.g. thumbnails etc).

Currently Explorer includes a "location" link on each dataset page that typically links to the dataset's JSON or YAML:
image

However, when a user clicks on this file, it simply downloads the JSON/YAML file rather than taking the user to the data itself. It would be far more useful if this link took users to the location of the data on S3 using the "https://data.dea.ga.gov.au/" browser. 

For example, for this dataset:
* https://explorer-aws.dea.ga.gov.au/products/ga_ls8c_ard_3/datasets/b7bb12c7-f8fd-4ddc-965d-3d957bb423e7#locations
* https://dea-public-data.s3.ap-southeast-2.amazonaws.com/baseline/ga_ls8c_ard_3/094/081/2013/05/17/ga_ls8c_ard_3-0-0_094081_2013-05-17_final.stac-item.json

You could be taken to this location:
* https://data.dea.ga.gov.au/?prefix=baseline/ga_ls8c_ard_3/094/081/2013/05/17/

A challenge here would be making this work across our different Explorers (e.g. dev and prod). I suspect it won't be suitable for the NCI Explorer as there's no way to browse those files without direct access to the NCI.

Ariana-B:
As per issue #480 

Config should now include `SHOW_DATASET_LOCATION`, which is a dict that maps an s3 bucket to its browseable url. For now, that is just `dea-public-data` mapping to `data.dea.ga.gov.au`. The first 'location' link for the dataset links to the browseable folder, and all offsets to their respective links from that page.

Added test to `test_resolved_uris` to verify functionality, including that other links or s3 buckets are unaffected by this change. 

Also took the chance to clean up thumbnail work a little: made filter code more pythonic, and updated css to make image width unset so that image ratio is conserved, allowing for proper overlay even if the dataset is irregularly-shaped.
Ex:
image
vs
image

----
:books: Documentation preview :books:: https://datacube-explorer--488.org.readthedocs.build/en/488/

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,6 +225,8 @@ You can alter default [Flask](http://flask.pocoo.org/docs/1.0/config/) or
     default_map_zoom = 3
     default_map_center = [-26.2756326, 134.9387844]
 
+    # S3 buckets for which data browser url should be returned
+    SHOW_DATA_LOCATION = { "dea-public-data": "data.dea.ga.gov.au" }
 
 [Sentry](https://sentry.io/) error reporting is supported by adding a `SENTRY_CONFIG` section.
 See [their documentation](https://docs.sentry.io/clients/python/integrations/flask/#settings).
--- a/cubedash/_filters.py
+++ b/cubedash/_filters.py
@@ -125,13 +125,8 @@ def _dataset_file_paths(dataset: Dataset):
 @bp.app_template_filter("dataset_thumbnail_url")
 def _dataset_thumbnail_url(dataset: Dataset):
     file_paths = _dataset_file_paths(dataset)
-    if "thumbnail:nbart" in file_paths:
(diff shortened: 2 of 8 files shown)


robbibt (approved):
This looks fantastic to me, thanks @Ariana-B! I don't have a good way to test this at the moment so can do a more thorough review/suggest any changes once it's live on an Explorer.

omad (approved):
Groovy!

PR #479, opened 2022-11-15
Fix tests (temp)

Ariana-B:
Put back digitalearthau testing plugins for now so that tests can run, until they can be properly cleaned up in core

----
:books: Documentation preview :books:: https://datacube-explorer--479.org.readthedocs.build/en/479/

PR #490, opened 2022-11-21
Replace Datacube Explorer Database Test Infrastructure

omad:
Continuing on from work started by @Ariana-B in #479 and myself in https://github.com/opendatacube/odc-tools/pull/524https://github.com/opendatacube/odc-tools/pull/524

This PR completely redoes the way database tests work for Datacube Explorer.

## Temporary Docker based PostgreSQL server

- If Docker is available, a temporary PostGIS server is run for the duration of the tests. (Pytest Fixture `postgresql_server`, scope = module)
    - If the database fixture detects it is running within Docker, and that ODC Database access is configured already, it does nothing, allowing the existing testing infrastructure using docker to continue working.
- An ODC Database is created and destroyed per test module, by a new fixture called  `odc_db` of scope `module`.  It creates a new database within the postgres server, named after the test module, and configures ODC to access it via environment variables and temporary configuration files.
- A new fixture called `odc_test_db`, of `module` scope, initialises the ODC database structure within the temporary database. It also removes the database after the module is finished.

## Test fixture data loading
Previously, test data was loaded a little bit all over the place, using a function which was copied into multiple test modules.

This has been replaced by a new fixture called `auto_odc_db`  of `module` scope, which looks for module variables (named `METADATA_TYPES`, `PRODUCTS`, and `DATASETS`), and loads them into the current database. This allows each test module to specify exactly which data it expects to be available.

## Benefits

In combination, this new setup:

- Allows the running of all the integration tests without any database setup, assuming `docker` is available.
- Allows running any test individually, with the correct fixtures being used to setup the database and test data.
- Simplifies and makes explicit which test data is used by which test module.
- Makes it possible to run multiple instances of `pytest` simultaneously (due to unique docker postgres instances), and multiple test modules simultaneously using `pytest-xdist`, due to unique databases being created per module.

These new fixtures have been implemented in a new `cubedash.testutils.database` module, which is loaded as a `pytest` plugin. My intention, once proven, is to move this test module into `datacube-core` to be available for all downstream ODC based software.

----
:books: Documentation preview :books:: https://datacube-explorer--490.org.readthedocs.build/en/490/


integration_tests/test_dataset_listing.py
```diff
@@ -8,13 +8,13 @@
 from cubedash._utils import DEFAULT_PLATFORM_END_DATE, query_to_search
 
 
-def test_parse_query_args(dea_index: Index):
+def test_parse_query_args(module_dea_index: Index):
```
Ariana-B: Shouldn't this be `odc_test_db`?

integration_tests/conftest.py
```diff
@@ -134,23 +100,21 @@ def dea_index_instance():
     return dea_index_instance
 
 
-module_index = index_fixture("module_db", scope="module")
-
-module_dea_index = dea_index_fixture("module_index", scope="module")
+module_dea_index_fixture = dea_index_fixture_factory("module_dea_index", scope="module")
```
Ariana-B: Is this fixture actually being used anywhere? Either way, there are several fixtures still using `module_dea_index` that will need to be updated

integration_tests/test_page_loads.py
```diff
@@ -32,7 +31,7 @@
 
 
 @pytest.fixture(scope="module", autouse=True)
-def auto_populate_index(populated_index: Index):
+def auto_populate_index(populated_index):
```
Ariana-B: should this be called `_auto_populate_index` for consistent naming conventions?

omad: Fixed by deleting :D

omad:
Switching back to Draft PR.

I haven't explained the testing changes anywhere near well enough, or documented the new pytest fixtures I've added. I just got excited that all the tests were passing.

pindge:
Please update this too https://github.com/opendatacube/datacube-explorer#contributing-to-integration-test

integration_tests/conftest.py
```diff
@@ -31,135 +21,31 @@
...
     )
     return store
 
 
-@pytest.fixture()
-def summariser(summary_store: SummaryStore):
-    return summary_store._summariser
-
-
 @pytest.fixture(autouse=True, scope="session")
```
Ariana-B, line 49: Can probably get rid of this too

PR #500, opened 2022-11-29
Implement stac extensions functionality

Ariana-B:
Implemented stac extension functionality for query, fields, sortby, and limited functionality for filter. 

- [Query extension](https://github.com/stac-api-extensions/query): bound to item search endpoint; all operators supported. If multiple properties are provided they are treated as a logical AND.
- [Fields extension](https://github.com/stac-api-extensions/fields): bound to item search and features endpoints, used with POST requests. `id`, `type`, `geometry`, `bbox`, `links`, `assets`, `properties.datetime`, and `stac_version` are always included for the returned entities to be valid STAC items.
- [Sort extension](https://github.com/stac-api-extensions/sort): bound to item search and features endpoints, following the POST JSON syntax.
- [Filter extension](https://github.com/stac-api-extensions/filter): bound to item search endpoint, supporting CQL2 JSON syntax and the following basic CQL2 operators: `AND`, `OR`, `=`, `>`, `<`, `>=`, `<=`, `<>`, `IS NULL`. 
 
Nested fields should be provided using dot notation (ex: `properties.datetime`).

Also cleaned up the `integration_tests/data` folder by moving sample dataset files into `data/datasets`, and updated the tests that use them accordingly.

----
:books: Documentation preview :books:: https://datacube-explorer--500.org.readthedocs.build/en/500/


omad (approved):
Thanks for cleaning up the test data some more.

I suspect that because this is performing filtering in Python land, it won't be fast enough for large queries, but it's a nice simple implementation, so lets try it out first.

Looks good.

PR #505, opened 2022-12-05
Make field search capability more discoverable

### LINKED ISSUE #491: Make field search capability more discoverable
I was trying to navigate explorer for a dataset of a particular maturity, and was doing it based on my knowledge of when datasets are processed. Then I remembered, explorer actually lets you search by fields - and it took me a minute to work out where to go.

To make this more discoverable, how about we add href links to the fields that point to the `/datasets` page, with that search field pre-filled out. You could even have a little search icon mouse over.

image
image

Ariana-B:
Made field search capability more discoverable as per #491. Also took the opportunity to remove some unnecessary imports and clean up dataset fields grid styling a little.

- search icon on hover
- text wrap for long values
image

Auto-determine bounds for 'time' and 'creation_time' to be 1 day when they are not provided, included when they are not provided as a range, as occurs in this use case. Also make dataset maturity search case-insensitive to avoid confusions due to capitalisation in the label.

----
:books: Documentation preview :books:: https://datacube-explorer--505.org.readthedocs.build/en/505/


pindge:
is there UI changes? could u take a screenshot in the description

pindge (approved):

cubedash/_pages.py
```diff
@@ -178,18 +178,39 @@ def search_page(
...
                 search_time.begin or product_summary.time_earliest,
                 search_time.end or product_summary.time_latest + timedelta(days=1),
             )
-    # The URL time range always trumps args.
-    if time_range:
-        query["time"] = time_range
+        query["time"] = search_time
+
+    # same logic as with 'time'
+    if "creation_time" in query:
```
pindge, line 202: does this include? https://github.com/opendatacube/datacube-core/issues/1295

Ariana-B: As far as I can tell, the search query string only has it as `creation_time`
ex: https://explorer.dev.dea.ga.gov.au/products/ga_s2am_ard_3/datasets?creation_time-begin=2022-08-04&creation_time-end=2022-12-08&time-begin=2016-09-21&time-end=2022-12-08 

I haven't changed anything related to core

pindge: cool

pindge: do you think you can set another integration test for this query?based off the below test?

https://github.com/opendatacube/datacube-explorer/blob/develop/integration_tests/test_page_loads.py#L515-L522

Ariana-B: Done!

PR #506, opened 2022-12-05
Improve image overlay

Ariana-B:
Determine image overlay bounds based on dataset bounds instead of extent for better overlap

image

----
:books: Documentation preview :books:: https://datacube-explorer--506.org.readthedocs.build/en/506/


robbibt (approved):
Looks fantastic! I'll do another closer review when it's live on Dev Explorer to see how it handles some weirder edge cases, but I think this should clear up most of the plotting issues for weird shaped geoms. Thanks @Ariana-B !

PR #509, opened 2022-12-14
fix css

Ariana-B:
Removed erroneous lines from css that were causing fields to be hidden

----
:books: Documentation preview :books:: https://datacube-explorer--509.org.readthedocs.build/en/509/


omad (approved):

omad:
I think the tests are failing because of an updated dependency. We should be pinning them here, since explorer is an application, not a library.

PR #512, opened 2023-02-01
update shapely version pin

Ariana-B:
Pin shapely >= 2.0.0 to be compatible with datacube-core

----
:books: Documentation preview :books:: https://datacube-explorer--512.org.readthedocs.build/en/512/


omad (approved):
Epic! ;-)

PR #515, opened 2023-02-08
Add basic /queryables endpoints

### LINKED ISSUE #513: STAC API Filter requires implementing  `/queryables` endpoints
I'm experimenting with some [STAC filter queries](https://github.com/stac-api-extensions/filter) for some DEA Products, and having trouble getting them to work. While reading the spec, I've noticed that explorer doesn't implement the [Queryables](https://github.com/stac-api-extensions/filter#queryables) endpoints, and it should as part of it's Filter implementation.

The implementation guidelines for the spec say:
> [start by] defining a static Queryables schema with no queryables advertised and the `additionalProperties` field set to `true`,

Ariana-B:
Add barebones `/queryables` endpoints for stac api catalog and collections as part of Filter implementation.

----
:books: Documentation preview :books:: https://datacube-explorer--515.org.readthedocs.build/en/515/

### DIFF
--- a/cubedash/_stac.py
+++ b/cubedash/_stac.py
@@ -836,6 +836,10 @@ def _stac_collection(collection: str) -> Collection:
                 rel="items",
                 target=url_for(".collection_items", collection=collection),
             ),
+            Link(
+                rel="http://www.opengis.net/def/rel/ogc/1.0/queryables",
+                target=url_for(".collection_queryables", collection=collection),
+            ),
         ]
     )
     if all_time_summary.timeline_dataset_counts:
@@ -931,6 +935,12 @@ def root():
                 media_type="application/json",
                 target=url_for(".stac_search"),
             ),
+            Link(
+                title="Queryables",
+                rel="http://www.opengis.net/def/rel/ogc/1.0/queryables",
+                media_type="application/json",
+                target=url_for(".queryables"),
+            ),
             # Individual Product Collections
             *(
                 Link(
@@ -1022,7 +1032,7 @@ def queryables():
     return _utils.as_json(
         {
             "$schema": "http://json-schema.org/draft-07/schema#",
-            "$id": "https://example.org/queryables",
+            "$id": flask.request.base_url,
             "type": "object",
             "title": "",
             "properties": {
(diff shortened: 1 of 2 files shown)


cubedash/_stac.py
```diff
@@ -1012,6 +1012,71 @@ def collections():
...
+def queryables():
+    """
+    Define what terms are available for use when writing filter expressions for the entire catalog
+    Part of basic CQL2 conformance for stac-api filter implementation.
+    See: https://github.com/stac-api-extensions/filter#queryables
+    """
+    return _utils.as_json(
+        {
+            "$schema": "http://json-schema.org/draft-07/schema#",
+            "$id": "https://example.org/queryables",
```
omad: The `$id` should be the current URL, not `example.org`. 

Although the response document is hard coded, in future it may change to include details on the queryable fields, and I'd expect that to be different per explorer instance.

cubedash/_stac.py
```diff
@@ -1012,6 +1012,71 @@ def collections():
...
+            "title": "",
+            "properties": {
+                "id": {
+                    "title": "Item ID",
+                    "description": "Item identifier",
+                    "$ref": "https://schemas.stacspec.org/v1.0.0/item-spec/json-schema/"
+                    "item.json#/definitions/core/allOf/2/properties/id",
+                }
+            },
+        }
```
omad, line 1065: I think we want `"additionalProperties": true` in this response too.

cubedash/_stac.py
```diff
@@ -1012,6 +1012,71 @@ def collections():
...
+    try:
+        dataset_type = _model.STORE.get_dataset_type(collection)
+    except KeyError:
+        abort(404, f"Unknown collection {collection!r}")
+
+    collection_title = dataset_type.definition.get("description")
+    return _utils.as_json(
+        {
+            "$schema": "http://json-schema.org/draft-07/schema#",
+            "$id": "https://example.org/queryables",
```
omad: This `$id` should again be the current URL, including explorer host name and the `collection` id.

omad (changes requested):
Looks good bar a couple of little changes! Thanks.

cubedash/_stac.py
```diff
@@ -1012,6 +1012,71 @@ def collections():
...
+            "type": "object",
+            "title": "",
+            "properties": {
+                "id": {
+                    "title": "Item ID",
+                    "description": "Item identifier",
+                    "$ref": "https://schemas.stacspec.org/v1.0.0/item-spec/json-schema/"
+                    "item.json#/definitions/core/allOf/2/properties/id",
+                }
```
omad: Actually/also, according to the specification document, this should include all of the STAC Item fields. 

> It is required to implement both of these endpoints, but for a STAC API, this may simply be a static document of the STAC-specific fields. A basic Queryables definitions for STAC Items **should include at least the fields id, collection, geometry, and datetime**.

omad (approved):

PR #523, opened 2023-03-03
Parse NCI filepaths to THREDDS location

### LINKED ISSUE #520: NCI local filesystem indexed thumbnails are inaccessible from web
image
 
https://explorer.dea.ga.gov.au/products/ga_ls9c_ard_3/datasets/abe45d2b-a54f-468d-98ea-2ffb30656260

Perhaps a wrapper can be written for NCI to handle these `/g/data` files and translate them to the corresponding THREDDS location? Not sure if the paths are always consistently 1:1 from `/g/data` and THREDDS, though. 🤷

|  |  |
|--|--|
| NCI location | file:///g/data/xu18/ga/ga_ls9c_ard_3/105/068/2022/10/06/ga_ls9c_nbart_3-2-1_105068_2022-10-06_final_thumbnail.jpg |
| THREDDS location | https://dapds00.nci.org.au/thredds/fileServer/xu18/ga_ls9c_ard_3/105/068/2022/10/06/ga_ls9c_nbart_3-2-1_105068_2022-10-06_final_thumbnail.jpg |
| Catalog | https://dapds00.nci.org.au/thredds/catalog/xu18/ga_ls9c_ard_3/105/068/2022/10/06/catalog.html |

---

This seems to be at least possible in principle, because that seems to be done for AWS - mapping from `S3://dea-public-data/...` -> `https://dea-public-data.s3.ap-southeast-2.amazonaws.com/...`
image

Perhaps, while we're at it we could replace the location link with the THREDDS ones too:
image

Looks like it's handled here in the code: https://github.com/opendatacube/datacube-explorer/blob/e09884af0b398d30f4e4c4494f9a808146dbec95/cubedash/_utils.py#L148-L182

Ariana-B:
As per #520 

----
📚 Documentation preview 📚: https://datacube-explorer--523.org.readthedocs.build/en/523/


dunkgray (approved):
looks ok, though I haven't worked on this code.

omad (changes requested):
Hi Ariana,

Can you please make this an opt-in/configurable option. Explorer isn't only used by DEA, and this change would be most unexpected in other environments. :)

Cheers

mpaget:
> an opt-in/configurable option

Yes, I expect this would be good to have configurable by a deployment admin. Sharing some ideas, if helpful.

If URL converting regexs or functions were imported from `settings.env.py` we'd need to consider the security implications and do some careful vetting. I'm not sure of the effort or appetite to look at it from that perspective. A transparent in-code solution would be easier.

As a deployment patch perhaps? Perhaps `settings.env.py` could reference an `as_external_url` function that can be imported by explorer, else the "s3" default is used. The custom `as_external_url` function would need to be uploaded to the host running explorer. This may require an update to the explorer chart too.

whatnick:
Could we please update this to latest develop to avoid potential `merge with rebase` issues, especially if global settings are being changed. This will affect the main file where the timezone setting was recently added.

omad:
Feel free to re-open, but I suspect this is no longer particularly useful, or applicable.

PR #540, opened 2023-11-10
Dockerfile: clean /code directory

pjonsson:
The /code directory in the image
is 38M because it contains .git,
.docker, and various other
configuration files.

Clean out the directory completely
to make the image slightly smaller.


pjonsson:
Readthedocs failure fixed in #541

omad (approved):
Thanks!

pjonsson:
@omad I don't understand how things are wired up for the tests, but it's possible this change broke the integration tests by removing `/code/.docker/create_db.sh`. Sorry for that.

The integration tests seem to already be broken on Werkzeug version though, so I have tried to fix that in #542.

PR #541, opened 2023-11-10
readthedocs: update config to current format

pjonsson:
The image tag has been renamed to os and
the python version goes under build.tools.python.
   
The support for system_packages has been dropped,
for more details, see:
    
https://blog.readthedocs.com/drop-support-system-packages/


omad (approved):
Thank you!

Issue #546, opened 2023-11-19
Fix JSON schema validation issues in tests


pjonsson:
Commit e243e975fa1b4 disabled the tests that were failing on JSON schema validation issues. The tests should be fixed, and that commit reverted after they have been fixed.

PR #548, opened 2023-11-19
pre-commit-config: do not reformat schemas

pjonsson:
Reformatting will give spurious differences
after running the update.sh script.

----
:books: Documentation preview :books:: https://datacube-explorer--548.org.readthedocs.build/en/548/

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,18 +49,20 @@ repos:
       - id: check-docstring-first # Avoid common error of code before docstring.
       - id: check-json # Check json file syntax
         # (Exclude Typescript config: it uses json5 extensions)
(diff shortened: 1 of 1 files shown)


pjonsson:
I'm not sure how to test this without doing an actual update and see what the precommit bot does, so I might have missed something. Considering #546, I don't want to change things too much.

omad (approved):

omad:
Sorry if I'm explaining things you already know...

Pre-commit should be installed locally, and then `pre-commit install` should be run for within each repo using it.

After doing that, it's possible to do a "test run" using `pre-commit run --all-files`, which is AFAIK what the pre-commit service does.

pjonsson:
I know very little about best practices for interpreted languages, thank you for explaining!

PR #555, opened 2023-11-21
docker-compose: add healthcheck for postgres

pjonsson:
This makes "docker compose up --wait" return
when the database can be connected to through
the hostname.

----
:books: Documentation preview :books:: https://datacube-explorer--555.org.readthedocs.build/en/555/


omad (approved):
Handy, thanks.

PR #556, opened 2023-11-24
CI: fix docpreview permissions

pjonsson:
According to

https://stackoverflow.com/questions/74957218/what-is-the-difference-between-pull-request-and-pull-request-target-event-in-git

there is something with permissions of tokens
that differ between pull_request and pull_request_target, so switch to pull_request_target which is
what the action documentation suggests.

### DIFF
--- a/.github/workflows/doc.yaml
+++ b/.github/workflows/doc.yaml
@@ -2,7 +2,7 @@
 name: Docs QA
 
 on:
-  pull_request:
+  pull_request_target:
     paths:
       - "**"
 


pjonsson:
A similar change works in the datacube-ows repository (the documentation-trigger will only start to work for outsiders once it is merged to main).

omad (approved):
Thanks, lets see how it goes.

To be honest, I didn't even realise Explorer _had_ a docs site!

PR #561, opened 2023-12-03
Dockerfile: build psycopg2 in separate container

pjonsson:
This avoids having to install development libraries in the production container, which makes the image ~300M smaller.

----
:books: Documentation preview :books:: https://datacube-explorer--561.org.readthedocs.build/en/561/


omad (approved):
Nice!

PR #564, opened 2023-12-07
Update to flask 3

pjonsson:
Remove the before_first_request decorator since it is deprecated in Flask 2.2.

Remove the check for schema compatibility,
generate.py already performs that check,
and conftest.py ensures the tables are
freshly created.

I don't know anything about Prometheus so I only know the CI tests pass,
I have no idea if Prometheus works as intended.

----
:books: Documentation preview :books:: https://datacube-explorer--564.org.readthedocs.build/en/564/


omad (approved):
Sounds good to me, thanks.

PR #570, opened 2024-01-11
Improve stac api item search extensions

Ariana-B:
Remedy some oversights from my previous work on stac api extensions:
- support `fields`, `sortby`, and `filter` as GET request query parameters
- remove conformance classes that weren't actually implemented (and don't need to be)
- make extension handling logic more in line with the extension specifications

This is a band-aid fix; the entire stac api implementation needs to be reworked. 

----
📚 Documentation preview 📚: https://datacube-explorer--570.org.readthedocs.build/en/570/


omad:
Merged in another PR

PR #578, opened 2024-03-04
Fix pagination with pystac-client >= 0.7.4

### LINKED ISSUE #575: Explorer STAC API search issue: only returning max of 20 items
A user on LinkedIn and @alexgleith have encountered a possible bug in our Explorer STAC search API ([see link here](https://www.linkedin.com/feed/update/urn:li:ugcPost:7163413218147725313?commentUrn=urn%3Ali%3Acomment%3A%28ugcPost%3A7163413218147725313%2C7163449042763440128%29&replyUrn=urn%3Ali%3Acomment%3A%28ugcPost%3A7163413218147725313%2C7163653885583187968%29&dashCommentUrn=urn%3Ali%3Afsd_comment%3A%287163449042763440128%2Curn%3Ali%3AugcPost%3A7163413218147725313%29&dashReplyUrn=urn%3Ali%3Afsd_comment%3A%287163653885583187968%2Curn%3Ali%3AugcPost%3A7163413218147725313%29)).
 
If you do a super simple query of DEA's Sentinel-2 data from December 2023 to Feb 2024, you only get back data up to January 17, despite the data definitely existing:
```
import pystac_client, odc.stac

client = pystac_client.Client.open("https://explorer.sandbox.dea.ga.gov.au/stac")

# Search for items in the collection
collections = ["ga_s2am_ard_3", "ga_s2bm_ard_3"]
query = client.search(
    collections=collections,
    bbox=[146.04, -34.30, 146.05, -34.28],
    datetime="2023-12-01/2024-02-28",
)

# Search the STAC catalog for all items matching the query
[i.properties["datetime"] for i in query.get_items()]
```
image

It seems that by default, the query is only returning the first 20 items from the query. To get any extra data, the user has to manually provide a high limit, e.g.:
```
query = client.search(
    collections=collections,
    bbox=[146.04, -34.30, 146.05, -34.28],
    datetime="2023-12-01/2024-02-28",
    limit=1000,
)
```
image

This isn't typical behavior for STAC loading: normally when using `pystac.client()` it will automatically follow "next" page links to provide the user with all datasets matching their query - the user definitely isn't limited to a tiny amount like 20.
 
It looks to me that Explorer might be using the `DEFAULT_PAGE_SIZE` of 20 to define the absolute limit of datasets returned. This doesn't appear to follow the correct STAC API approach ([see Slack conversation here](https://opendatacube.slack.com/archives/CAREN05K3/p1707948879911369) and [STAC API docs here](https://github.com/radiantearth/stac-api-spec/tree/release/v1.0.0/item-search#query-parameter-table)). I can see this line which seems like it might the source of the issue - it seems to use `DEFAULT_PAGE_SIZE` if no limit is provided:
https://github.com/opendatacube/datacube-explorer/blob/3cdcf98a7394eb85566609a4f9cbf6f22009b722/cubedash/_stac.py#L433
 
As it is, I think the current functionality is confusing to our users - they will naturally expect to get back all items matching their query (at least up to some sensibly high limit, definitely not 20), and only getting back half the time series is pretty unexpected.

Ariana-B:
Pystac-client 0.7.4 removed default `limit`, which exposed an error with our pagination.

- Format `collections` in link so that it remains a comma-separated list -- otherwise we get duplicate `collections` keys that `TypeConversionDict` won't catch
- Have `next` link method match method of original request, rather than defaulting to `GET` unless `intersects` value was specified
- Add title and type to next link

----
📚 Documentation preview 📚: https://datacube-explorer--578.org.readthedocs.build/en/578/

### DIFF
--- a/cubedash/_stac.py
+++ b/cubedash/_stac.py
@@ -338,7 +338,7 @@ def _build_properties(d: DocReader):
 # Search arguments
 
 
-def _array_arg(arg: str | list, expect_type=str, expect_size=None) -> List:
+def _array_arg(arg: Union[str, list], expect_type=str, expect_size=None) -> List:
     """
     Parse an argument that should be a simple list.
     """
@@ -347,6 +347,8 @@ def _array_arg(arg: str | list, expect_type=str, expect_size=None) -> List:
 
     # Make invalid arguments loud. The default ValueError behaviour is to quietly forget the param.
     try:
+        if not isinstance(arg, str):
+            raise ValueError
         arg = arg.strip()
         # Legacy json-like format. This is what sat-api seems to do too.
         if arg.startswith("["):
--- a/cubedash/_stac.py
+++ b/cubedash/_stac.py
@@ -2,7 +2,6 @@ import json
 import logging
 import uuid
 from datetime import datetime, time as dt_time, timedelta
-from functools import partial
 from typing import Callable, Dict, List, Optional, Sequence, Tuple, Union
 
 import flask
@@ -19,7 +18,7 @@ from pystac import Catalog, Collection, Extent, ItemCollection, Link, STACObject
 from shapely.geometry import shape
 from shapely.geometry.base import BaseGeometry
 from toolz import dicttoolz
-from werkzeug.datastructures import TypeConversionDict
+from werkzeug.datastructures import MultiDict
 from werkzeug.exceptions import BadRequest, HTTPException
 
 from cubedash.summary._stores import DatasetItem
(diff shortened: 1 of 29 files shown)


cubedash/_stac.py
```diff
@@ -338,7 +338,7 @@ def _build_properties(d: DocReader):
 # Search arguments
 
 
-def _array_arg(arg: str, expect_type=str, expect_size=None) -> List:
+def _array_arg(arg: Union[str, list], expect_type=str, expect_size=None) -> List:
```
SpacemanPaul: Can we elaborate further?   `list[str]` instead of just `list`?

SpacemanPaul (approved):
I'm not overly familiar with this part of Explorer so I might be missing something, but the changes look good.

robbibt:
Thanks so much @Ariana-B! Can you let me know when this has been rolled out so I can test this out on a range of typical STAC queries?

PR #586, opened 2024-04-06
Update to latest postgis 13 version

pjonsson:
I don't know to what extent explorer is affected by database bugs, but this bumps to the most recent version of the same major.

Also throw in a commit that avoids the download progress bar from docker compose since that floods the CI logs with noise.

----
📚 Documentation preview 📚: https://datacube-explorer--586.org.readthedocs.build/en/586/


omad:
Thanks!

PR #587, opened 2024-04-08
Revert previous auto-modernise commits

jeremyh:
Some changes by shed have broken the semantics of sqlalchemy queries (see https://github.com/opendatacube/datacube-explorer/issues/581)

- Revert the original, offending commit that breaks result iteration.
- Also update our usage of `== None`, which has a better equivalent in modern sqlalchemy (and was also broken by it)
- Move from shed to ruff. Most of our newer repositories used ruff anyway, so it seems a good reason to switch.

I haven't retested the original issues, this is a just reverting some clear errors.

**Note** that ruff has made some changes of its own to match its black-like formatting, but they're minor

----
📚 Documentation preview 📚: https://datacube-explorer--587.org.readthedocs.build/en/587/

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,28 +1,11 @@
 ---
 repos:
-  # Normalise all Python code. (Black + isort + pyupgrade + autoflake)
-  - repo: https://github.com/Zac-HD/shed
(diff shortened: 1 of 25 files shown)


omad (approved):
Thanks!

PR #588, opened 2024-04-08
Dockerfile: remove git after use

pjonsson:
I shifted some things around to simplify
before pushing, but missed that the check
on ENVIRONMENT ended up before
the declaration of the argument.

Even the deployment container needs git
since setuptools_scm requires it, so
remove git and git-man after the
installation has completed.

This avoids CVE-2018-1000021.

----
📚 Documentation preview 📚: https://datacube-explorer--588.org.readthedocs.build/en/588/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,18 +28,17 @@ ENV DEBIAN_FRONTEND=noninteractive \
     PYTHONFAULTHANDLER=1
 
 # Apt installation
+# git: required by setuptools_scm.
 RUN apt-get update && \
     apt-get upgrade -y && \
     apt-get install -y --no-install-recommends \
+      git \
       # For Psycopg2
       libpq5 \
       tini \
       postgresql-client \
       python3-pip \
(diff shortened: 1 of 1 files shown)


pjonsson:
I have to run now, will look at the failure tomorrow. And just in case, can someone try to retrigger the failing job and see if it works itself out?

pjonsson:
The issue was that testing started despite the image not being built properly.

That touches the parts I'm already changing, so will fix error code propagation after this is merged.

omad (approved):
Holy moly, this Dockerfile is crazy complicated. 

Thanks

pjonsson:
@omad Fix for the build error masking is in #596.

And I agree with your assessment, a build/app container setup must be >50% of the usages of Docker, and it's still really complicated if one cares about build performance as well as the final result. Personally I've gone monorepo and a single Dockerfile with different build-args to produce all images, and while it's sometimes a bit more difficult to do some things that way, at least I only have to do them once and I can sleep at night since I don't have to worry I forgot to update some repository.

PR #589, opened 2024-04-08
Dockerfile: disable pip cache in deployment

pjonsson:
Tell pip to not write the cache when
in the deployment image.

Also disable the version check on
the remaining uses of pip, this saves
some time since it avoids a HTTP
request on startup.

----
📚 Documentation preview 📚: https://datacube-explorer--589.org.readthedocs.build/en/589/


omad (approved):
Thanks

PR #591, opened 2024-04-11
Rework stac api extension logic

### LINKED ISSUE #536: STAC API returning error when a specific field has no value
STAC specification says: "No error must be returned if a specified field has no value for it in the catalog. For example, if the field "properties.eo:cloud_cover" is specified but there is no cloud cover value for an Item, a successful HTTP response must be returned and the Item entities will not contain that field.

DEAfrica user raised an issuee that a Server Error is returned when a request includes an attribte in "fields" that is not defined for a collection, e.g. `curl 'https://explorer.digitalearth.africa/stac/search' -H 'content-type: application/json' --data-raw '{"limit": 10, "bbox": [-20.0, 10.0, -19.0, 11.0], "datetime": "2023-01-02T00:00:00.000001Z/2023-02-01T00:00:00Z", "collections": ["s1_rtc"], "fields": {"include": ["properties.sat:absolute_orb"]}}'`
This impacts search over multiple collections that have different attributes defined.

Looks like this might relates to how the fields search is handled.

### LINKED ISSUE #579: STAC search pagination is applied before query filtering
Regarding STAC API extensions, the README says:

> The STAC endpoint implements the [query](https://github.com/stac-api-extensions/query), [filter](https://github.com/stac-api-extensions/filter), [fields](https://github.com/stac-api-extensions/fields), and [sort](https://github.com/stac-api-extensions/sort) extensions, all of which are bound to the search endpoint as used with POST requests, with fields and sort additionally bound to the features endpoint.
> 
> Fields contained in the item properties must be prefixed with properties., ex properties.dea:dataset_maturity.

From this information, I figured that to limit my results by maximum cloud cover, I need to `POST` to `/stac/search` with a body like this:

```json
{
    "query": {
        "properties.eo:cloud_cover": {
            "lte": 10
        }
    }
}
```

But the `numberMatched` of the returned result is the same as when I omit this `query`. This leaves me wondering that maybe my JSON is wrong? Adding an example to the README explaining how to use the query API would really help to rule this out, as right now I'm not sure whether it's a problem with my request or my data. I tried various variants of the request (e.g. adding `"gte": 0`), but none seems to do anything. But also my data _does_ have the `eo:cloud_cover` attribute, which also _is_ included in the `properties` of the returned `features`, so I don't really see a problem there either...

Ariana-B:
Add parameters to `SummaryStore.search_items` to enable factoring the `filter` and `sortby` extensions into the database query, thus moving the logic out of `_handle_search_request` and `search_stac_items` in `_stac.py`, where it was previously causing issues when results were paginated, as highlighted in #579.

Also took the opportunity to fix logic that was missing or out of line with the STAC API spec, such as reported in #536 and initially attempted in #570.

The STAC API extension specs are sometimes unclear or ambiguous around expected behaviour, so I have chosen to interpret them in the ways that seem the most reasonable and consistent. Most notably,
-  `sortby` will ignore any undefined properties, 
- both `sortby` and `filter` will recognise any potential variant of a property name (stac name (ex: `created`), eo3 name (ex: `odc:processing_datetime`), and metadata type search field name (ex: `creation_time`)) as syntactically valid for that property.

----
📚 Documentation preview 📚: https://datacube-explorer--591.org.readthedocs.build/en/591/


omad (approved):
I think this looks okay, lets just release it and try it...

PR #597, opened 2024-05-11
Add dockerignore

pjonsson:
This avoids bringing in various
local files into the docker build
context, so the docker cache will
be tainted less often, improving
the performance of image rebuilds.

If the docker cache gets in the way,
one can still use docker build --no-cache.

----
📚 Documentation preview 📚: https://datacube-explorer--597.org.readthedocs.build/en/597/


omad (approved):
Thanks, wise.

PR #601, opened 2024-05-16
Restrict search to indexed fields

Ariana-B:
As per title; try and help improve search latency by disallowing search by fields that aren't indexed

----
📚 Documentation preview 📚: https://datacube-explorer--601.org.readthedocs.build/en/601/


cubedash/_filters.py
```diff
@@ -11,7 +11,7 @@
 import flask
 import pytz
 from datacube.index.fields import Field
-from datacube.model import Dataset, DatasetType, Range
+from datacube.model import Dataset, Product, Range
```
SpacemanPaul, line 14: I obviously approve, but this is breaking the documentation build - the documentation build must be using an ancient version of core.

SpacemanPaul (approved):
Looks good to me - not sure why that test is failing though.

PR #614, opened 2024-06-28
Use Flask application factory 

Ariana-B:
Rather than creating the Flask app module-level across `_model.py` and `_pages.py`, use the preferred application factory pattern to instantiate.
This highlighted an issue around how the default timezone is determined. Having `CUBEDASH_DEFAULT_TIMEZONE` be provided via both an environment variable and the app config seems like overkill, and which version is used - as well as the default value - appears to be inconsistent. Timezone configuration would likely be best revisited and properly reworked in another PR, but for now I suggest having it configurable only via the app config.

----
📚 Documentation preview 📚: https://datacube-explorer--614.org.readthedocs.build/en/614/

### DIFF
--- a/cubedash/_model.py
+++ b/cubedash/_model.py
@@ -16,6 +16,7 @@ from flask_themer import Themer
 # pylint: disable=import-error
 from sentry_sdk.integrations.flask import FlaskIntegration
 from shapely.geometry import MultiPolygon
+from werkzeug.exceptions import HTTPException
 
 # Fix up URL Scheme handling using this
 # from https://stackoverflow.com/questions/23347387/x-forwarded-proto-and-flask
@@ -25,6 +26,8 @@ from cubedash.summary import SummaryStore, TimePeriodOverview
 from cubedash.summary._extents import RegionInfo
 from cubedash.summary._stores import ProductSummary
 from cubedash.summary._summarise import DEFAULT_TIMEZONE
+
+from . import _utils as utils
 
 try:
     from ._version import version as __version__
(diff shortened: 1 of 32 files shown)


whatnick:
Please fix merge conflicts, not sure why the CI has paused.

cubedash/_model.py
```diff
@@ -57,70 +61,120 @@
         # release="myapp@1.0.0",
     )
 
+cache = Cache()
```
omad, line 64: In the future we should try and wrap up the initialisation of the sentry_sdk and SummaryStore as [flask extensions](https://flask.palletsprojects.com/en/3.0.x/extensiondev/), so that they can get configuration from the app, instead of loading files themselves or via environment variables.

cubedash/_monitoring.py
```diff
@@ -23,7 +24,7 @@ def init_app_monitoring():
 
     _INITIALISED = True
 
-    @_model.app.before_request
+    @current_app.before_request
```
omad: I don't think this will work if monitoring is enabled, current_app won't be anything real when this is called.

I think instead, app should be passed as an argument to the `init_app_monitoring()`

cubedash/summary/_summarise.py
```diff
@@ -29,7 +29,19 @@
 
 _NEWER_SQLALCHEMY = not sqlalchemy.__version__.startswith("1.3")
 
-DEFAULT_TIMEZONE = environ.get("CUBEDASH_DEFAULT_TIMEZONE", "Australia/Darwin")
+# Get default timezone via CUBEDASH_SETTINGS specified config file if it exists,
+# otherwise default to Australia/Darwin
+default_timezone = "Australia/Darwin"
+settings_file = os.environ.get("CUBEDASH_SETTINGS", "settings.env.py")
```
omad, line 35: Ewww, this isn't good at all. We'll have to do something better. But if it works, it can be in a later PR.

omad (changes requested):
Looking good, I think one thing to change now, and some more for later (because they're harder, and this is still a big improvement).

cubedash/summary/_summarise.py
Ariana-B: The only alternative I could think of was to manually create a Flask Config object and then use its `from_pyfile` method, but neither solution seemed good so I went with this one to avoid importing flask here for this sole purpose. It should hopefully work, but I think the default timezone logic is in dire need of a larger rework anyway.
If there's a less painful interim solution, I'd be happy to implement it.

omad: This is fine for now, it's ugly, but I like that it's contained in a single file that we can clean up in future.

omad (approved):

PR #615, opened 2024-07-02
More explicit error for removed Query extension

### LINKED ISSUE #611: Add error or more direct warning to deprecated "query" extension
The most recent version of Explorer no longer supports the "query" syntax (in favour of STAC's recommend "filter" syntax).

Although this is mentioned in a new warning from PySTAC, this is not very descriptive and doesn't explain that users should use "filter" instead:
image

This change risks users running into "silent failure" issues where they think that "query" is being applied when it is not (and getting back unfiltered data accidentally).

To avoid this, we should add either a hard error (forcing users to adapt their code to use "filter" instead), or a more descrptive warning message advising users to swap to "filter".

Ariana-B:
While the STAC API `Query` extension was previously implemented, it suffered from the same flaws in logic as the other API extensions, which meant that while it would return correct results, they risked being oddly distributed when pagination was involved. Since it is recommended to implement the `Filter` extension instead of `Query`, and since correctly re-implementing `Query` was not trivial, it was dropped in favour of `Filter`.
As mentioned in #611, `pystac_client` will warn users if they try to use `query`, but will still return results without issue. The api itself will similarly quietly ignore `query` if provided.
To avoid any silent failure issues down the line, return a 400 error if the request includes a `query` argument.

----
📚 Documentation preview 📚: https://datacube-explorer--615.org.readthedocs.build/en/615/

### DIFF
--- a/cubedash/_stac.py
+++ b/cubedash/_stac.py
@@ -559,6 +559,13 @@ def _handle_search_request(
                     "Only 'id', 'collection', and Item properties can be used to sort results.",
                 )
 
+    # Make sure users know that the query extension isn't implemented
+    if request_args.get("query") is not None:
+        abort(
+            400,
+            "The Query extension is no longer supported. Please use the Filter extension instead.",
+        )
+
     filter_lang = request_args.get("filter-lang", default=None, type=str)
     filter_cql = request_args.get("filter", default=None, type=_filter_arg)
     filter_crs = request_args.get("filter-crs", default=None)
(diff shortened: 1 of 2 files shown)


robbibt:
Hey @Ariana-B, this looks great. Is there any way I can test this on the STAC endpoint before merging? If not, I'll just test it out after we merge and release a new Explorer version

robbibt (approved):
This looks great to me - let's merge this and I'll double check it once it's live 🙂

PR #616, opened 2024-07-08
Update robots.txt default content

### LINKED ISSUE #522: Keep Away Well-behaved robots

Ariana-B:
As recommended, also disallow `/dataset/*` pages in default robots.txt to avoid bots hitting individual datasets.

----
📚 Documentation preview 📚: https://datacube-explorer--616.org.readthedocs.build/en/616/

### DIFF
--- a/cubedash/_pages.py
+++ b/cubedash/_pages.py
@@ -40,9 +40,7 @@ _DEFAULT_GROUP_NAME = "Other Products"
 
 _DEFAULT_ARRIVALS_DAYS = 14
 
-_ROBOTS_TXT_DEFAULT = (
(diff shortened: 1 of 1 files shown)


omad (changes requested):
Makes sense, thank you. 

Can you please make it a triple quoted string though, it'll be much more readable.

omad (approved):
Thank you!

PR #620, opened 2024-07-25
Dockerfile: fixate GDAL version

pjonsson:
Building from gdal-latest will make
the explorer 2.12.3 image unbuildable
when gdal-latest suddenly contains
breaking changes.

Fixate the image version to keep old
releases of explorer buildable even
in the future.

----
📚 Documentation preview 📚: https://datacube-explorer--620.org.readthedocs.build/en/620/


pjonsson:
@Ariana-B I try to stay up to date with releases, but there are a lot of things to keep up with, so when I fall behind it's valuable for me if the old releases remain buildable.

Ariana-B (approved):
I had not thought of that - good catch, and thank you!

PR #634, opened 2024-11-15
Add PostGIS support

Ariana-B:
Create the `ExplorerIndex` class to enable PostGIS index support.

----
📚 Documentation preview 📚: https://datacube-explorer--634.org.readthedocs.build/en/634/


cubedash/_utils.py
```diff
@@ -105,13 +99,44 @@ def expects_eo3_metadata_type(md: MetadataType) -> bool:
...
+    """
+    # If EO3+Stac formats, there's already has a plain 'datetime' field,
+    # So we can use it directly.
+    if expects_eo3_metadata_type(md_type):
+        props = jsonb_doc_expression(md_type)["properties"]
+
+        # .... but in newer Stac, datetime is optional.
+        # .... in which case we fall back to the start time.
+        #      (which I think makes more sense in large ranges than a calculated center time)
+        return (
```
SpacemanPaul, line 127: I totally agree about start time making more sense than centre time for large ranges.

cubedash/_utils.py
```diff
@@ -426,7 +451,7 @@ def dataset_created(dataset: Dataset) -> Optional[datetime]:
             _LOG.warning(
                 "invalid_dataset.creation_dt", dataset_id=dataset.id, value=value
             )
-
+    # should we fall back to the indexed time, like in _dataset_creation_expression?
     return None
```
SpacemanPaul, line 458: Falling back to indexed time makes sense to me tbh.

cubedash/_utils.py
```diff
@@ -435,6 +460,7 @@ def center_time_from_metadata(dataset: Dataset) -> datetime:
     This function shares the same logic as
     https://github.com/opendatacube/datacube-explorer/blob/4afa0dbbb51d541f377c479e7edb914bdb62aef9/cubedash/summary/_extents.py#L481-L505
     """
+    # seems to be a misleading name...
```
SpacemanPaul, line 467: IKR - what is this function?

cubedash/index/postgis/_api.py
```diff
@@ -0,0 +1,977 @@
...
+            )
+
+    def upsert_product_record(self, product_name: str, fields):
+        # Dear future reader. This section used to use an 'UPSERT' statement (as in,
+        # insert, on_conflict...) and while this works, it triggers the sequence
+        # `product_id_seq` to increment as part of the check for insertion. This
+        # is bad because there's only 32 k values in the sequence and we have run out
+        # a couple of times! So, It appears that this update-else-insert must be done
+        # in two transactions...
+        with self.index._active_connection() as conn:
```
SpacemanPaul: Doing in two SQL statements makes sense, but could (and probably should) still be done in a single transaction:

Just use `self.index.transaction()` instead of `self.index._active_connection()`

cubedash/index/postgis/_api.py
```diff
@@ -0,0 +1,977 @@
...
+    def product_time_overview(self, product_id: int):
+        with self.index._active_connection() as conn:
+            return conn.execute(
+                select(
+                    func.min(DatasetSpatial.center_time),
+                    func.max(DatasetSpatial.center_time),
+                    func.count(),
+                ).where(DatasetSpatial.product_ref == product_id)
+            ).fetchone()
+            # return conn.execute(
```
SpacemanPaul: Do we need this commented out code?

cubedash/index/api.py
```diff
@@ -0,0 +1,193 @@
...
+from datacube.index import Index
+from datacube.model import Dataset, MetadataType, Product, Range
+from sqlalchemy.sql import ColumnElement
+
+
+class EmptyDbError(Exception):
+    pass
+
+
+class ExplorerAbstractIndex(ABC):
```
SpacemanPaul, line 15: Eventually it would be nice to use this (and the equivalent abstract class in OWS) as a checklist of functionality to be brought into the core index API.

It would therefore be nice if all these methods could have clear docstrings.  Probably not worth holding up the 1.9 release over though.

cubedash/testutils/database.py
```diff
@@ -100,7 +102,9 @@ def odc_db(postgresql_server, tmp_path_factory, request):
             tmp_path_factory.mktemp("odc") / "test_datacube.conf"
         )
         config = configparser.ConfigParser()
```
SpacemanPaul, line 104: With the new config API, it feels like there should be a simpler way to handle test config than dynamically generating a temporary ini file.

SpacemanPaul (approved):
Mostly looks good.  Feel free to address any of my comments that can addressed quickly, but simply adding a comment or a GH issue to come back to them later would also be an acceptable response.

(And obviously get the GHA checks passing)

cubedash/index/postgis/_api.py
Ariana-B: ~~Nope, there were just so many of them scattered around that I've missed a few!~~
Actually, this was an alternative query that I thought should work, but didn't

cubedash/_utils.py
Ariana-B: This is an old comment, the method just got moved out of `_extents.py`

cubedash/testutils/database.py
Ariana-B: Yeah, this whole setup is copypasta from odc-tools and needs to be revisited

PR #648, opened 2025-03-23
schemas: update to STAC 1.1.0

pjonsson:
----
📚 Documentation preview 📚: https://datacube-explorer--648.org.readthedocs.build/en/648/


pjonsson:
There are 3 tests still failing that I couldn't get to work. I suggest merging this so the CI becomes green, and then someone who knows more about explorer and its testsuite can take a look at it.

Edit: a log of the failures can be found in https://github.com/opendatacube/datacube-explorer/actions/runs/14019751866/job/39250105214.

Why do the tests print a million lines of ` == Index Call == get_dataset: 1.4252662658691406` on stderr?

omad (approved):
Thanks Peter!!!

PR #651, opened 2025-03-24
tests: fix robots.txt test

pjonsson:
The text was updated in #642
without updating the test,
so update the test.

----
📚 Documentation preview 📚: https://datacube-explorer--651.org.readthedocs.build/en/651/


omad (approved):
Whoops, and thanks.

Such a fragile test... but, I guess it's better than nothing.

PR #654, opened 2025-03-30
Update to Python 3.10 syntax

pjonsson:
Use modern syntax for types,
and import things from collection
instead of typing when possible.

----
📚 Documentation preview 📚: https://datacube-explorer--654.org.readthedocs.build/en/654/


omad (approved):
That's a lot of cleaning 🧹! Thanks.

PR #655, opened 2025-03-30
Fix utcnow() warning

pjonsson:
Use datetime.now(timezone.utc) instead.

----
📚 Documentation preview 📚: https://datacube-explorer--655.org.readthedocs.build/en/655/


omad (approved):
👍🏻

PR #658, opened 2025-03-31
Fix some grid_spec warnings

pjonsson:
Make a private function that
extracts the same product information
that was previously gotten from
the now deprecated grid_spec,
and then use that helper instead
of grid_spec.

The helper should be removed once
there are suitable replacements
in some part of the opendatacube-API.

----
📚 Documentation preview 📚: https://datacube-explorer--658.org.readthedocs.build/en/658/


pjonsson:
This fixes 550 warnings that `test_page_load` triggers, 220 warnings triggered by `test_stac`, and a bunch of other warnings.

After this PR, it's clear that the remaining GridSpec warnings are caused by `test_summarise_data.py::test_computed_regions_match_those_summarised[default]`.

pjonsson:
Making this a draft since https://github.com/opendatacube/datacube-core/pull/1783 revealed the resolution can be a scalar.

If the grid_spec property remains and is no longer deprecated, I think it would be better to use that than replicate the same functionality in explorer.

PR #665, opened 2025-04-10
api: align argument with implementations

pjonsson:
The two implementations have no stars,
so remove the stars on the abstract
method as well.

----
📚 Documentation preview 📚: https://datacube-explorer--665.org.readthedocs.build/en/665/


pjonsson:
I think Paul is quite busy with OWS so adding omad and Ariana as reviewers. I have too many things in flight right now so I might have mixed things up, but I think Ariana was the last one who touched the lines for the changes in api.

Ariana-B (approved):
Thanks for this!
I have been the main person looking after explorer lately, so happy to be the primary reviewer on any PRs.

PR #671, opened 2025-04-20
Import names from defining module

pjonsson:
Import names from the module where
they are defined instead of some
module where they are imported.

----
📚 Documentation preview 📚: https://datacube-explorer--671.org.readthedocs.build/en/671/

### DIFF
--- a/cubedash/_filters.py
+++ b/cubedash/_filters.py
@@ -9,13 +9,13 @@ from datetime import datetime, timezone
 from urllib.parse import quote_plus
 
 import flask
+import orjson
 import pytz
 from datacube.index.fields import Field
 from datacube.model import Dataset, Product, Range
 from dateutil import tz
 from flask import Blueprint
 from markupsafe import Markup, escape
-from orjson import orjson
 from shapely.geometry import MultiPolygon
 
 from . import _model, _utils
--- a/cubedash/_pages.py
+++ b/cubedash/_pages.py
@@ -3,12 +3,12 @@ import re
 from datetime import datetime, timedelta, timezone
 
 import datacube
-import dateutil.parser
 import flask
 import structlog
 from datacube.model import Product, Range
 from datacube.scripts.dataset import build_dataset_info
 from dateutil import tz
+from dateutil.parser import ParserError
 from flask import (
     Blueprint,
     abort,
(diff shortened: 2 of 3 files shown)


cubedash/_pages.py
```diff
@@ -3,12 +3,12 @@
 from datetime import datetime, timedelta, timezone
 
 import datacube
-import dateutil.parser
 import flask
 import structlog
 from datacube.model import Product, Range
 from datacube.scripts.dataset import build_dataset_info
 from dateutil import tz
+from dateutil.parser._parser import ParserError
```
omad:
```suggestion
from dateutil.parser import ParserError
```

See https://dateutil.readthedocs.io/en/stable/parser.html#warnings-and-exceptions

cubedash/_filters.py
```diff
@@ -15,7 +15,7 @@
 from dateutil import tz
 from flask import Blueprint
 from markupsafe import Markup, escape
-from orjson import orjson
+from orjson.orjson import dumps
```
omad:
```suggestion
import orjson
```
Follow the examples in the [orjson docs](https://pypi.org/project/orjson/), and also the common usage of the `json` builtin module.

cubedash/_filters.py
```diff
@@ -64,7 +64,7 @@ def _dataset_label(dataset):
 
 @bp.app_template_filter("torapidjson")
 def _fast_tojson(obj):
-    return Markup(orjson.dumps(obj).decode("utf-8"))
+    return Markup(dumps(obj).decode("utf-8"))
```
omad:
```suggestion
    return Markup(orjson.dumps(obj).decode("utf-8"))
```

omad (changes requested):

cubedash/_pages.py
pjonsson: Fixed.

cubedash/_filters.py
pjonsson: Fixed.

cubedash/_filters.py
pjonsson: Fixed.

omad (approved):
Thanks!

PR #672, opened 2025-04-20
docs/conf: remove hashbang

pjonsson:
This file is never executed, so
remove the hashbang.

----
📚 Documentation preview 📚: https://datacube-explorer--672.org.readthedocs.build/en/672/

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,5 +1,3 @@
-#!/usr/bin/env python
-#
 # datacube_ows documentation build configuration file, created by
 # sphinx-quickstart on Tue Jul  9 22:26:36 2013.
 #


omad (changes requested):
This file is never directly executed, so I'd prefer to remove this line, and it doesn't need to be executable. It's probably an artefact of an old version of `sphinx-quickstart`.

[`conf.py` in the official sphinx docs](https://github.com/sphinx-doc/sphinx/blob/master/doc/conf.py#L1-L8) doesn't include one.

pjonsson:
> This file is never directly executed, so I'd prefer to remove this line

Even better, fixed!

omad (approved):

PR #673, opened 2025-04-26
dependabot: update requirements.txt 

pjonsson:
This aligns with the behavior
of Dependabot in datacube-core.

----
📚 Documentation preview 📚: https://datacube-explorer--673.org.readthedocs.build/en/673/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,9 +4,7 @@ updates:
   directory: "/"
   schedule:
     interval: "daily"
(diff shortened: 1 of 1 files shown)


pjonsson:
I don't think this caused the pre-commit failure, but not sure how to retrigger that job.

Edit: probably caused by Github errors, here's an old ticket with the same issue: https://github.com/pre-commit-ci/issues/issues/220

PR #674, opened 2025-04-26
CI: remove unused permissions

pjonsson:
This gets flagged by CodeQL scanning.

----
📚 Documentation preview 📚: https://datacube-explorer--674.org.readthedocs.build/en/674/

### DIFF
--- a/.github/workflows/deployment_test.yaml
+++ b/.github/workflows/deployment_test.yaml
@@ -16,6 +16,8 @@ on: # yamllint disable-line rule:truthy
 env:
   IMAGE_NAME: opendatacube/explorer
 
+permissions: {}
+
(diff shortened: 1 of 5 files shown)


pjonsson:
I don't know about the publish_pypi workflow ~~but I'm guessing that doesn't write or do anything requiring permissions at Github~~, For the rest of the files I believe they're fine without permissions if the CI passes.

Edit: @omad solved PyPi in https://github.com/opendatacube/datacube-core/pull/1715, so copy that solution verbatim. And use the same permissions for the scan workflow as the example at https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/uploading-a-sarif-file-to-github#example-workflow-for-sarif-files-generated-outside-of-a-repository

omad (approved):

PR #706, opened 2025-06-27
pytest: output test runtimes

pjonsson:
Enable the pytest 8.4 feature to display
test times.

Beyond finding the slow tests more easily,
this can also be used as an indication
of CI machines suffering from some kind
temporary hickup when failing.

----
📚 Documentation preview 📚: https://datacube-explorer--706.org.readthedocs.build/en/706/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,7 @@ warn_unused_ignores = true
 
 [tool.pytest.ini_options]
 addopts = "--doctest-modules"
+console_output_style = "times"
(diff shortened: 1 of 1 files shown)


omad:
I'd be inclined to leave the listing of the slowest tests there as well, I like it being highlighted and easier to find. And I think it's per test instead of per file. 

Generally, I think there's no harm having extra output in ci logs.

omad (approved):
:D Thank you

PR #727, opened 2025-07-21
Convert setup.py to pyproject.toml

pjonsson:
The same files are included before/after
when doing "python3 -m build" on
my machine.

----
📚 Documentation preview 📚: https://datacube-explorer--727.org.readthedocs.build/en/727/


omad (approved):
Thanks!

Issue #708, opened 2025-07-02
Implement STAC-API Collection Search


caitlinadams:
When I use pystac-client to search for collections in our dev STAC endpoint, I receive the following warnings:
```
MissingLink: No link with rel='data' could be found on this Client.
UserWarning: Server does not conform to COLLECTION_SEARCH. Filtering will be performed client-side where only bbox, datetime, and q arguments are supported
```

Is it possible to implement the [collection-search extension](https://github.com/stac-api-extensions/collection-search) to avoid this warning being triggered?

Example code:
```
from pystac_client import Client
catalog = "https://explorer.dev.dea.ga.gov.au/stac"

stac_client = Client.open(catalog)
results = stac_client.collection_search(q="ga_s1")
```

whatnick:
Should this be closed via #730 , we are looking for this feature and would appreciate a 3.0.1 release.

SpacemanPaul:
@whatnick - 3.0.1 was released today (August 11th)

PR #730, opened 2025-07-23
Implement STAC API Collection Search

Ariana-B:
As per #708, implement basic collection search functionality.

----
📚 Documentation preview 📚: https://datacube-explorer--730.org.readthedocs.build/en/730/


cubedash/index/postgis/_api.py
```diff
@@ -377,6 +377,90 @@ def product_time_summary(self, product_id: int, start_day, period):
                 )
             )
 
+    @override
+    def collection_cols(self):
+        """Get all columns necessary for creating a Collection"""
```
pjonsson: Can't the documentation string be put on the abstract method instead, and all implementing classes automatically gets the documentation string? (That is how my PyCharm behaves with the documentation in some other code base, and I haven't changed any configuration.)

cubedash/index/api.py
```diff
@@ -101,6 +101,20 @@ def put_summary(self, product_id: int, start_day, period, summary_row: dict): ..
     @abstractmethod
     def product_summary_cols(self, product_name: str): ...
 
+    @abstractmethod
+    def collection_cols(self): ...
```
pjonsson: I think this method returns `Select`.

cubedash/index/api.py
```diff
@@ -101,6 +101,20 @@ def put_summary(self, product_id: int, start_day, period, summary_row: dict): ..
...
+    @abstractmethod
+    def collections_search_query(
+        self,
+        limit: int,
+        offset: int,
+        name: str | None = None,
+        bbox: tuple[float, float, float, float] | None = None,
+        time: tuple[datetime, datetime] | None = None,
+        q: list[str] | None = None,
+    ): ...
```
pjonsson: What is the return type for this method?

cubedash/_stac.py
```diff
@@ -655,6 +702,49 @@ def next_page_url(next_offset):
...
+    q = request_args.get("q", default=None, type=partial(_array_arg, expect_type=str))
+
+    limit = request_args.get("limit", default=get_default_limit(), type=int)
+
+    offset = request_args.get("_o", default=0, type=int)
+
+    check_page_limit(limit)
+
+    if bbox is not None and len(bbox) != 4:
+        abort(400, "Expected bbox of size 4. [min lon, min lat, max long, max lat]")
```
pjonsson: First component is called `lon` and third component is called `long`, I'm guessing they should be called the same?

The type annotations for various bboxes in this code declare it as a tuple. Is the on-wire format a list?

cubedash/_stac.py
```diff
@@ -655,6 +702,49 @@ def next_page_url(next_offset):
...
+
+    q = request_args.get("q", default=None, type=partial(_array_arg, expect_type=str))
+
+    limit = request_args.get("limit", default=get_default_limit(), type=int)
+
+    offset = request_args.get("_o", default=0, type=int)
+
+    check_page_limit(limit)
+
+    if bbox is not None and len(bbox) != 4:
```
pjonsson: If this check fails, will the server return a 400 page and no code after the `abort()` call gets executed? If so, the check should come right after bbox is defined so the server doesn't have to do a bunch of work that gets thrown away.

cubedash/_stac.py
```diff
@@ -860,76 +938,62 @@ def search_stac_items(
...
-        ),
+    returned = collections[:limit]
+    there_are_more = len(collections) == limit + 1
+
+    count_matching = len(
+        list(
+            _model.STORE.search_collections(
+                time=time,
+                bbox=bbox,
+                q=q,
```
pjonsson: I think Ruff will reflow this if you remove the trailing comma after `q`.

cubedash/_stac.py
```diff
@@ -979,6 +1043,28 @@ def root_catalog():
 # ENDPOINTS
 ##########################
 
+CONFORMANCE_CLASSES = [
+    "http://www.opengis.net/spec/ogcapi-features-1/1.0/conf/core",
+    "http://www.opengis.net/spec/ogcapi-features-1/1.0/conf/geojson",
+    "http://www.opengis.net/spec/ogcapi-features-1/1.0/conf/oas30",
+    "https://api.stacspec.org/v1.0.0-rc.1/core",
```
pjonsson: Not for this PR, but hasn't the 1.0.0 been out in the wild for a few years by now, so it should be safe to conform to that API?

cubedash/_stac.py
```diff
@@ -1187,7 +1263,13 @@ def collection(collection: str):
     """
     Overview of a WFS Collection (a datacube product)
     """
-    return _stac_response(_stac_collection(collection))
+    # return _stac_response(_stac_collection(collection))
+    try:
+        _model.STORE.get_product(collection)
```
pjonsson, line 1257: Why does this line call `get_product`, and the return-statement calls `get_collection`?

cubedash/_stac.py
```diff
@@ -1187,7 +1263,13 @@ def collection(collection: str):
     """
     Overview of a WFS Collection (a datacube product)
     """
-    return _stac_response(_stac_collection(collection))
+    # return _stac_response(_stac_collection(collection))
```
pjonsson: Leftover debug code?

cubedash/index/postgis/_api.py
```diff
@@ -377,6 +377,90 @@ def product_time_summary(self, product_id: int, start_day, period):
...
+                indexed=False,
+                offset=["metadata", "title"],
+            )
+
+            description = SimpleDocField(
+                name="description",
+                description="product description",
+                alchemy_column=collection.c.definition,
+                indexed=False,
+                offset=["description"],
```
pjonsson: The type of offset is `tuple | Sequence[tuple]`, so this should be `offset=("description",),`, and same with the offset in the other `SimpleDocField` instantiations.

cubedash/summary/_stores.py
```diff
@@ -211,6 +212,40 @@ def as_geojson(self):
...
+    time_latest: datetime | None
+    footprint_geometry: Geometry | None
+    footprint_crs: str | None
+
+    @property
+    def footprint_wgs84(self) -> MultiPolygon | None:
+        if not self.footprint_geometry:
+            return None
+        if not self.footprint_crs:
+            _LOG.warning(f"Geometry without a crs for {self}", stacklevel=2)
```
pjonsson: Isn't `name` sufficient to identify the faulty item, or does the entire object need to be logged?

cubedash/summary/_stores.py
```diff
@@ -1624,6 +1678,29 @@ def _summary_to_row(
     )
 
 
+def _row_to_collection(res, grouping_timezone=default_timezone):
```
pjonsson: Can this function get some type annotations?

cubedash/summary/_stores.py
```diff
@@ -1624,6 +1678,29 @@ def _summary_to_row(
...
+        else None,
+        footprint_geometry=(
+            None
+            if res.footprint_geometry is None
+            else geo_shape.to_shape(res.footprint_geometry)
+        ),
+        footprint_crs=(
+            None
+            if res.footprint_geometry is None or res.footprint_geometry.srid == -1
+            else "EPSG:{}".format(res.footprint_geometry.srid)
```
pjonsson, line 1701: Can use an f-string for this?

cubedash/summary/_stores.py
```diff
@@ -1166,6 +1201,25 @@ def search_items(
                 odc_dataset=(self.index.datasets.get(r.id) if full_dataset else None),
             )
 
+    def search_collections(
+        self,
+        name: str | None = None,
+        time: Tuple[datetime, datetime] | None = None,
```
pjonsson: Is there some SQLAlchemy `Tuple` in scope here, or can the parameters use `tuple`?

cubedash/_stac.py
Ariana-B: I suppose so; I think it's only declared as a tuple elsewhere so as to specify the expected length. I can expand those type annotations.

cubedash/_stac.py
Ariana-B: Because `get_product` returns a datacube `Product`, which does not have all the necessary Collection fields, and here it only serves to check the existence of the specified collection. We could use `get_collection` from the get-go and just check if returns None, but I thought `get_product` may be preferable in the first instance since it's a cached method.

cubedash/index/postgis/_api.py
Ariana-B: Documentation and type annotation improvements need to be done not just for what I've implemented in this PR but for all other methods as well. I think that should be handled in a separate PR.

cubedash/_stac.py
pjonsson: My thought was whether the user was passing in the bbox using parentheses or brackets, and that the error message should use the same notation.

I think it makes sense to use a tuple internally to have the length checked as you say.

Ariana-B: In a query, bbox is represented as something like `bbox=160.6,-55.95,-170,-25.89`. Our logic for parsing array-type arguments allows for it to be in square brackets (as it would be in json representation) but that's not the expected format.

cubedash/index/postgis/_api.py
SpacemanPaul: I understand where you are coming from, but I think I'm with Peter on this one.  Better to annotate the functions you are touching in this PR in this PR, and leave the functions you aren't otherwise touching in this PR to another PR.  While typehints in functions otherwise untouched make the PR bigger and haerder to review, typehints on the functions you ARE touching make it easier for reviewers to understand what's going on in this PR.

SpacemanPaul (approved):
I agree with Peter that (in general) it's better to typehint untyped functions touched by a PR within that PR, as it makes it easier for reviewers to understand what's going on.

Actual implementation looks solid.

PR #731, opened 2025-07-23
CI: remove sleep statements

pjonsson:
I added a health check to
docker-compose.yaml at
the end of 2023, start
using that to decide if
the container is up instead
of sleeping for an arbitrary
number of seconds.

----
📚 Documentation preview 📚: https://datacube-explorer--731.org.readthedocs.build/en/731/


omad (approved):
wonderful

PR #734, opened 2025-07-25
Remove field type mismatch check

Ariana-B:
Checking that all dataset fields are of the correct type in `_find_product_fixed_metadata` has been behind several errors as we work through some 1.9 updates, and may not even be necessary (at all, or in this location). Removing.

----
📚 Documentation preview 📚: https://datacube-explorer--734.org.readthedocs.build/en/734/

### DIFF
--- a/cubedash/summary/_stores.py
+++ b/cubedash/summary/_stores.py
@@ -547,19 +547,19 @@ class SummaryStore:
             if field.type_name in simple_field_types and name in first_dataset_fields
         ]
 
-        # I don't think this check is necessary?
+        # TODO: remove if confirmed unnecessary, or make less stringent
         # Give a friendlier error message when a product doesn't match the dataset.
-        for name, field in candidate_fields:
-            sample_value = first_dataset_fields[name]
-            expected_types = simple_field_types[field.type_name]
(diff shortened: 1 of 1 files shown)


cubedash/summary/_stores.py
```diff
@@ -547,19 +547,19 @@ def _find_product_fixed_metadata(
             if field.type_name in simple_field_types and name in first_dataset_fields
         ]
 
-        # I don't think this check is necessary?
+        # TODO: remove if confirmed unnecessary, or make less stringent
```
pjonsson: I think it's better to delete the code+comments instead so we don't have to look at some commented out code for months/years.

If it turns out this PR is wrong, `git revert` will  be equally happy to restore deleted lines as it is to uncomment lines.

SpacemanPaul (approved):
This test is basically confirming database integrity.  The case being checked for can only arise from:

- a bug in the indexing code from core,
- someone doing something naughty with the "allow-unsafe" option (i.e. telling core "I know what I'm doing" when they didn't). or
- someone manually messing with the database,

Either way, it's not clear why it would be this code's responsibility to check this.

I think it's safe to remove fully.

PR #736, opened 2025-07-25
tests: Fix output explosion in test logs. Run timing test last

pjonsson:
The timing decorator modifies global
state which cannot be undone, so run
that test last.

This created an enormous explosion of log output, completely drowning out everything else. 

----
📚 Documentation preview 📚: https://datacube-explorer--736.org.readthedocs.build/en/736/


SpacemanPaul (approved):
This seems like it should not be necessary, but the way `_monitoring.init_app_monitoring` is written does make it very hard to "undo".

PR #738, opened 2025-07-26
CI: quote version number

pjonsson:
This will be interpreted as
3.1 without quotes.

----
📚 Documentation preview 📚: https://datacube-explorer--738.org.readthedocs.build/en/738/


SpacemanPaul (approved):
If I had a dollar for every time we've made this mistake, I'd have enough for a beer.

PR #749, opened 2025-07-29
CI: cache docker layers

pjonsson:
Switch to the same image build setup as used by datacube-core to speed up image building.

Leave the Docker workflow as-is for now, it looks like it should be simplified before being changed.

----
📚 Documentation preview 📚: https://datacube-explorer--749.org.readthedocs.build/en/749/

### DIFF
--- a/.github/workflows/deployment_test.yaml
+++ b/.github/workflows/deployment_test.yaml
@@ -14,7 +14,7 @@ on: # yamllint disable-line rule:truthy
       - '**'
 
 env:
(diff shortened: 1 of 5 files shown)


pjonsson:
This resembles a similar PR in datacube-ows, see my comment there for a screenshot of the job summary: https://github.com/opendatacube/datacube-ows/pull/1234#issuecomment-3132390156

SpacemanPaul (approved):

PR #750, opened 2025-07-29
readthedocs: use uv for building

pjonsson:
This uses uv.lock for the dependencies,
which eliminates the need for an extra
requirements.txt in docs/.

Remove the automatic updates from
Dependabot without any replacement
since the uv eco-system in Dependabot
currently seems too buggy to be used
for our setup.

----
📚 Documentation preview 📚: https://datacube-explorer--750.org.readthedocs.build/en/750/


pjonsson:
This also fixes the faulty version in the upper left corner of the documentation: 
screenshot of faulty version in other PR

SpacemanPaul (approved):

PR #755, opened 2025-07-31
Dockerfile: use uv for build

pjonsson:
Use uv for building the Docker image.

This also shuffles things around a bit to improve the Docker cache hit rate. When building locally, this will also cache uv itself as well as the deb packages.

----
📚 Documentation preview 📚: https://datacube-explorer--755.org.readthedocs.build/en/755/

### DIFF
--- a/cubedash/testutils/database.py
+++ b/cubedash/testutils/database.py
@@ -4,7 +4,9 @@ import time
 from collections import Counter
 from pathlib import Path
 
-import docker
+# There is a docker directory in the root, skip sorting that import until
+# this repository uses a src layout.
+import docker  # isort: skip
 import psycopg2
 import psycopg2.extensions
 import pytest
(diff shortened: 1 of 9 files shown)


pjonsson:
The deployment image shrinks from 1.94 GB to 1.01 GB with these changes.

Not sure how coverage was working before since I don't find any traces of the package even before my conversion to use `pyproject.toml`, but I added the package in the first two commits in this PR.

cubedash/testutils/database.py
```diff
@@ -4,7 +4,9 @@
 from collections import Counter
 from pathlib import Path
 
-import docker
+# There is a docker directory in the root, skip sorting that import until
```
omad, line 7: Ugh 🤦🏻! Nice catch!

omad (approved):
Amazing!

PR #759, opened 2025-08-02
Dockerfile: remove proj specials

pjonsson:
The proj-stuff was an attempt to align
the Explorer and OWS images, but I
think installing an old proj when using
a modern GDAL image is a mistake.

----
📚 Documentation preview 📚: https://datacube-explorer--759.org.readthedocs.build/en/759/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,10 +29,8 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
         libhdf5-dev \
         libnetcdf-dev \
         libudunits2-dev \
-        libproj-dev \
         # For psycopg2.
         libpq-dev \
-        proj-bin \
         python3-dev
 
 ENV UV_COMPILE_BYTECODE=0 \
@@ -55,7 +53,6 @@ RUN --mount=type=cache,id=opendatacube-uv-cache,target=/root/.cache \
     uv sync --frozen --extra=deployment --no-install-project \
       --no-binary-package fiona \
       --no-binary-package netcdf4 \
-      --no-binary-package pyproj \
       --no-binary-package psycopg2 \
       --no-binary-package rasterio \
       --no-binary-package shapely


Dockerfile
```diff
@@ -55,18 +53,23 @@ RUN --mount=type=cache,id=opendatacube-uv-cache,target=/root/.cache \
...
+# the running image to unconditionally copy the files.
 RUN --mount=type=cache,id=opendatacube-uv-cache,target=/root/.cache \
     EXTRAS=$( ([ "$ENVIRONMENT" = "deployment" ] && echo "--extra=deployment --no-dev") || \
                  echo "--extra=test") \
-    && uv sync --frozen $EXTRAS --no-editable
+    && uv sync --frozen $EXTRAS --no-editable \
+    && ([ "$ENVIRONMENT" != "deployment" ] || \
+        (chmod 644 /usr/local/bin/uv* && \
+         echo "" > /usr/local/bin/uv && \
+         echo "" > /usr/local/bin/uvx))
```
omad: I have no idea what this binary truncation is supposed to be for. The comment doesn't help.

Is it even supposed to be here?

omad (changes requested):

Dockerfile
pjonsson: Beyond that the previous comment in the Dockerfile was not helpful, I'm not sure what your comment means.

I'm guessing you are wondering about some mechanical aspect of Docker or Docker images, so I'll answer some questions you didn't ask. Each `ADD/COPY/RUN` in the Dockerfile results in a layer, so `COPY --from=builder <files>` followed by  `RUN rm <files>` adds two layers to the image, the first layer adding a bunch of bytes and the second layer removing those bytes. There isn't a conditional copy in Docker, so instead we render the binaries harmless by truncating them to 0 bytes when building the deployment image, and then we always copy the "binaries" to the final image. This avoids creating the `RUN` layer at all, and the `COPY` layer becomes small for the deployment image since it is copying two 0 byte files.

I reworded the comment in the Dockerfile to try to convey the big picture of why we truncate the binaries.

pjonsson: These were two unrelated changes, so I've dropped this from this PR.

omad (approved):
Looks good to me! Thanks Peter

PR #758, opened 2025-08-01
Require urllib3 2.x

pjonsson:
This fixes CVE-2025-50181.

----
📚 Documentation preview 📚: https://datacube-explorer--758.org.readthedocs.build/en/758/

PR #761, opened 2025-08-03
Replace `requests-html` (unmaintained) with BeautifulSoup

omad:
If this works, I think it's a better approach than #758 

----
📚 Documentation preview 📚: https://datacube-explorer--761.org.readthedocs.build/en/761/


pjonsson:
> If this works, I think it's a better approach than #758

If you have enough time before the next release to get this working, I agree that this is a better choice and fully support this.

If you don't have enough time before the next release, I think #758 should be merged so we don't get a release image with a bunch of CVEs that could easily have been avoided.

PR #766, opened 2025-08-04
Dockerfile: avoid copying in uv

pjonsson:
Truncate the uv binaries to 0 bytes
for deployment image. This reduces
the image size by 46 MB.

This also removes the conditional
parts of the produced image, which
should improve Docker cache hit rates.

----
📚 Documentation preview 📚: https://datacube-explorer--766.org.readthedocs.build/en/766/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,10 +60,18 @@ RUN --mount=type=cache,id=opendatacube-uv-cache,target=/root/.cache \
 COPY --link . /build/
 
 ARG ENVIRONMENT=deployment
+# The deployment image should not have binaries that aid an attacker to get their
+# rootkit in place, and uv downloads over the network. There is no conditional
+# copy in Docker, so truncate the uv binaries to 0 bytes to render them harmless
+# in the resulting deployment image.
 RUN --mount=type=cache,id=opendatacube-uv-cache,target=/root/.cache \
     EXTRAS=$( ([ "$ENVIRONMENT" = "deployment" ] && echo "--extra=deployment --no-dev") || \
                  echo "--extra=test") \
-    && uv sync --frozen $EXTRAS --no-editable
+    && uv sync --frozen $EXTRAS --no-editable \
+    && ([ "$ENVIRONMENT" != "deployment" ] || \
+        (chmod 644 /usr/local/bin/uv* && \
+         echo "" > /usr/local/bin/uv && \
+         echo "" > /usr/local/bin/uvx))
 
 FROM base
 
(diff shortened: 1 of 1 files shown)


pjonsson:
Here's a screenshot from `dive` (CAUTION: DO NOT USE THE SNAP PACKAGES OF DIVE!), with the deployment image for develop on the left, and this PR on the right:
image
So the layers don't work exactly like I thought they did, but the net effect remains: truncating the binaries in the builder image before copying instead of copying+deleting in the deployment image reduces the image size by ~5%.

For reference, according to the times printed by Docker in the CI logs it takes 29.5s to export to docker image format, 13.5s to import ("load") from BuildKit to Docker, and 32.4s to prepare the build cache for export. Considering we are dealing with a ~1GB image, those numbers seem really high to me, but they are consistent with my recollection of my attempts of trying to load a saved image from the Github cache to get a speedup, and Docker's slow loading completely ruining the experience.

omad:
How about instead of doing this dance with truncating them in the builder, and then copying them, we just don't copy them into the released image at all?

Doesn't that achieve the same result, and is simpler and less confusing?

pjonsson:
> Doesn't that achieve the same result, and is simpler and less confusing?

It is simpler, but then the test image ends up without uv, which is now one of the essential development tools for ODC.

omad:
> > Doesn't that achieve the same result, and is simpler and less confusing?
> 
> It is simpler, but then the test image ends up without uv, which is now one of the essential development tools for ODC.

Aaah, okay. Now I'm starting to understand this whole mess! (I've also looked at the whole Dockerfile again).

Probably, we shouldn't be mixing the production and test Dockerfile into one file. It's confusing and complicated and messes up caching. 

Maybe there's better ways of compose Dockerfiles now, but, composability is absolutely not one of Docker's strengths.

Dockerfile
```diff
@@ -85,11 +93,6 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
```
omad:
```suggestion
# In the "deployment" build, these `uv` binaries will be 0 bytes.
# In the "test" build they're the actual `uv` tools.
```

omad (approved):

pjonsson:
> Probably, we shouldn't be mixing the production and test Dockerfile into one file. It's confusing and complicated and messes up caching.
> 
> Maybe there's better ways of compose Dockerfiles now, but, composability is absolutely not one of Docker's strengths.

There has been support for multistage docker files for quite a few years by now, so the natural way to solve this would be to have `FROM base AS deployment\n....`and then `FROM deployment AS test` and have the `COPY --from=builder /usr/local/uv* ...` in the test image.

Just like we had sleep statements in the CI until the other week, I'm just working with the current constraints of these repositories, long-term I see no reason to avoid multistage Docker files.

PR #767, opened 2025-08-04
index: fix AbstractIndex.Index type

pjonsson:
The class is never initialized with
Index = None, so make a constructor
so the type checker doesn't get
confused.

Also add some ignores and align the type
signatures between the base class
and the implementing classes.

----
📚 Documentation preview 📚: https://datacube-explorer--767.org.readthedocs.build/en/767/


cubedash/index/api.py
```diff
@@ -19,9 +19,14 @@ class EmptyDbError(Exception):
 
 
 class ExplorerAbstractIndex(ABC):
-    name: str = ""
-    index: Index = None
-    engine = None
+    def __init__(self, name: str, index: Index) -> None:
+        self.name = name
+        self.index = index
+        # There's no public api for sharing the existing engine (it's an implementation detail of the current index).
```
omad, line 25: I think this is fine. The ODC Core API should be improved to allow this sort of usage without using a protected attribute.

omad (approved):

cubedash/index/api.py
pjonsson: I just transferred the comment from the postgres constructor so I have no idea about the accuracy/validity/etc.

What you say does sound reasonable though, and without knowing any details for this particular case, I agree with you in general.

PR #769, opened 2025-08-04
Various type fixes

pjonsson:
Add type annotations/ignores so the files directly under cubedash pass the type checker, and verify that in the CI.

----
📚 Documentation preview 📚: https://datacube-explorer--769.org.readthedocs.build/en/769/

### DIFF
--- a/cubedash/logs.py
+++ b/cubedash/logs.py
@@ -6,7 +6,7 @@ from functools import partial
 from typing import BinaryIO
 
 import structlog
-from orjson import orjson
+from orjson.orjson import OPT_SORT_KEYS, dumps
 from structlog.types import EventDict, WrappedLogger
 from typing_extensions import override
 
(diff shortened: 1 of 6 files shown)


cubedash/logs.py
```diff
@@ -6,7 +6,7 @@
 from typing import BinaryIO
 
 import structlog
-from orjson import orjson
+from orjson.orjson import OPT_SORT_KEYS, dumps
```
omad, line 9: The orjson docs use `import orjson` and access functions and options from there. I think accessing everything directly on the `orjson` module is clearer. `dumps` is a very generic function name on it's own.

Example from the docs.
```python
import orjson, datetime, numpy
data = {
    "type": "job",
    "created_at": datetime.datetime(1970, 1, 1),
    "status": "🆗",
    "payload": numpy.array([[1, 2], [3, 4]]),
}
orjson.dumps(data, option=orjson.OPT_NAIVE_UTC | orjson.OPT_SERIALIZE_NUMPY)
```

omad (approved):
Look good, thanks.

Just a minor comment about `orjson`. I might raise a PR later to make it consistent everywhere. It's obviously far from critical :)

cubedash/logs.py
pjonsson: It's possible there is something with the environment in Explorer, but both MyPy and Pyright disagrees with the documentation when I run `uv run <typechecker> cubedash/logs.py`.

PR #775, opened 2025-08-05
Remove setuptools_scm workarounds

pjonsson:
These look like things that are masking
issues with packaging.

Coincidentally, this also seems to
reduce the "every file imports everything
else" that you notice when type checking
a single file.

----
📚 Documentation preview 📚: https://datacube-explorer--775.org.readthedocs.build/en/775/

### DIFF
--- a/cubedash/__init__.py
+++ b/cubedash/__init__.py
@@ -1,8 +1,4 @@
-try:
-    from ._version import version as __version__
-except ImportError:
-    __version__ = "Unknown/Not Installed"
-
 from ._model import create_app
+from ._version import version as __version__
 
 __all__ = ("create_app", "__version__")
--- a/Makefile
+++ b/Makefile
@@ -122,4 +122,4 @@ lint-docker: ## Run linting inside inside Docker
 
 test-docker: ## Run tests inside Docker
 	docker compose run --rm explorer \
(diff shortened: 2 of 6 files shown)


cubedash/__init__.py
```diff
@@ -1,8 +1,4 @@
-try:
```
pjonsson, line 1: I think these try-except are workarounds for not running an installed package. The docker image in CI is tested with installed packages*, and `uv run pytest` installs the package first, so this shouldn't be required any more.

*) The CI kept finding `cubedash` in `.` instead of using the installed version, so I changed import-mode of pytest, and pointed it to the integration_tests directory. I suspect these would be non-issues with a source-layout.

pjonsson:
Apparently everything breaks when trying to remove these things, so I'll get back to this later.

cubedash/__init__.py
omad: I've been thinking we should switch to a source-layout too. I think it's a more rigorous structure, with the only drawback being you can run `python` from the repo base and have it find the package source.

I think the days are long (loooong) past where that's a useful thing to want to do.

omad: I was caught out recently with this repo having a `docker/` directory, and it breaking `import docker`. :-(

PR #799, opened 2025-08-12
Dockerfile: stop installing psql

pjonsson:
Create the test database when starting
the postgis container instead of doing
it from the client side.

----
📚 Documentation preview 📚: https://datacube-explorer--799.org.readthedocs.build/en/799/

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -43,6 +43,7 @@ services:
       - "35433:35433"
     restart: always
     volumes:
+      - ./.docker/create_db.sh:/docker-entrypoint-initdb.d/create-pg-databases.sh
       - type: tmpfs
         target: /var/lib/postgresql/data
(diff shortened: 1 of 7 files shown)


docker-compose.yml
```diff
@@ -43,6 +43,7 @@ services:
       - "35433:35433"
     restart: always
     volumes:
+      - ./.docker/create_db.sh:/docker-entrypoint-initdb.d/create-pg-databases.sh
```
omad, line 48:
```suggestion
      # See https://hub.docker.com/_/postgres/#initialization-scripts
      # Create `opendatacube_test` database in addition to `opendatacube`
      - ./.docker/create_db.sh:/docker-entrypoint-initdb.d/create-pg-databases.sh
```

omad (approved):
Nice, I like less manual steps.

Maybe one day soon there'll be an arm64 postgis image to match the postgresql one. :(

PR #800, opened 2025-08-13
CI: split into multiple test jobs

pjonsson:
We get 20 concurrent CI runners, so this should reduce the time it takes to run the CI on a pull request.

The split makes a PR complete all the workflows in 12-14 minutes instead of 30-33 minutes. This is especially important if the intermittent "closed cursor" error happens, since a single job can be retriggered and the results are completed within 15 minutes.

----
📚 Documentation preview 📚: https://datacube-explorer--800.org.readthedocs.build/en/800/

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,10 +31,21 @@ concurrency:
 
 jobs:
   integration-tests:
-    timeout-minutes: 55
+    timeout-minutes: 30
     runs-on: ubuntu-latest
     permissions:
       id-token: write
+    strategy:
+      matrix:
+        # Tests take a long time, so split them over multiple CI jobs.
+        # Ignoring tests inverts the logic which makes it harder to understand,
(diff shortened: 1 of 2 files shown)


pjonsson:
> Additional details and impacted files
> ```diff
> @@             Coverage Diff             @@
> ##           develop     #800      +/-   ##
> ===========================================
> + Coverage    84.85%   84.95%   +0.09%     
> ===========================================
>   Files           34       34              
>   Lines         4227     4227              
>   Branches       532      532              
> ===========================================
> + Hits          3587     3591       +4     
> + Misses         453      446       -7     
> - Partials       187      190       +3     
> ```

No idea about this one, apparently we are now hitting the JSON dump in logs.py and for some reason we weren't doing that before. I have bugs to fix so I'm not going to investigate the root cause, I'll take the accidental win of increased coverage.

SpacemanPaul (approved):

PR #804, opened 2025-08-18
STAC: various type fixes

pjonsson:
----
📚 Documentation preview 📚: https://datacube-explorer--804.org.readthedocs.build/en/804/


SpacemanPaul (approved):
I think we might want to move a lot of this code to core fairly soon, but better type hints are always appreciated.

PR #810, opened 2025-08-20
ProductSummary: store time as tuple

pjonsson:
Combine the time_earliest and time_latest
fields in the ProductSummary, so the summary
will either have both an earliest and a latest
time, or no time at all, never just one
of them.

This fixes about a dozen errors from
the type checker.

----
📚 Documentation preview 📚: https://datacube-explorer--810.org.readthedocs.build/en/810/

### DIFF
--- a/cubedash/summary/_stores.py
+++ b/cubedash/summary/_stores.py
@@ -102,9 +102,8 @@ class GenerateResult(Enum):
 class ProductSummary:
     name: str
     dataset_count: int
-    # Null when dataset_count == 0
-    time_earliest: datetime | None
-    time_latest: datetime | None
+    # Tuple of (time_earliest, time_latest), None when dataset_count == 0.
+    full_time: tuple[datetime, datetime] | None
 
     source_products: list[str]
     derived_products: list[str]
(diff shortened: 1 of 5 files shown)


cubedash/summary/_stores.py
```diff
@@ -102,9 +102,8 @@ class GenerateResult(Enum):
 class ProductSummary:
     name: str
     dataset_count: int
-    # Null when dataset_count == 0
-    time_earliest: datetime | None
-    time_latest: datetime | None
+    # Tuple of (time_earliest, time_latest), None when dataset_count == 0.
+    full_time: tuple[datetime, datetime] | None
```
omad, line 106: I don't like this name, I don't follow sports, and yet that's what it makes me think of. 

My suggestions would be either `time_span` or `duration`.

pjonsson:
The detailed code coverage numbers are faulty for some reason, the real delta compared to develop is more or less +-0. The benefit of this PR is that it makes it clear that either there is a time range or there is None, not only one of them. As a consequence, that fixes a bunch of type checker complaints.

omad (approved):

cubedash/summary/_stores.py
pjonsson: Fixed in #818.

PR #826, opened 2025-09-02
Multiplatform/arch uv support in Docker

omad:
The Docker Image here was failing to build for me on an ARM Mac, because it was using a Linux ARM base image with an x64 binary of uv.

Install uv from a distroless docker image instead of downloading the binary directly. This is documented at [uv_docker] and is an easy way to support multiple platforms/architectures.

[uv_docker]: https://docs.astral.sh/uv/guides/integration/docker/#installing-uv

----
📚 Documentation preview 📚: https://datacube-explorer--826.org.readthedocs.build/en/826/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,9 @@ ENV LC_ALL=C.UTF-8 \
     PYTHONFAULTHANDLER=1 \
     PYTHONUNBUFFERED=1
 
+# This cannot be inlined below (e.g., COPY --from=...) because Dependabot does not support that syntax yet
+FROM ghcr.io/astral-sh/uv:0.8.15@sha256:a5727064a0de127bdb7c9d3c1383f3a9ac307d9f2d8a391edc7896c54289ced0 AS uv
+
 FROM base AS builder
 
 RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
@@ -39,7 +42,7 @@ ENV UV_COMPILE_BYTECODE=0 \
 
 WORKDIR /build
 
-COPY --from=ghcr.io/astral-sh/uv:0.8.14 /uv /uvx /bin/
+COPY --link --from=uv /uv /uvx /usr/local/bin/
 
 COPY --link pyproject.toml uv.lock /build/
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -41,9 +39,7 @@ ENV UV_COMPILE_BYTECODE=0 \
 
 WORKDIR /build
 
-ADD --checksum=sha256:5429c9b96cab65198c2e5bfe83e933329aa16303a0369d5beedc71785a4a2f36 --chown=root:root --chmod=644 --link $UV uv.tar.gz
-
-RUN tar xf uv.tar.gz -C /usr/local/bin --strip-components=1 --no-same-owner
+COPY --from=ghcr.io/astral-sh/uv:0.8.14 /uv /uvx /bin/
```
pjonsson: The copy should use `--link`, and there is a later step copying the binaries from `/usr/local/bin` so the binaries need to go in that directory.

On the security side, the `--from` doesn't contain a sha256 like the `FROM` at the top of this file, so if someone manages to switch where `0.8.14` is pointing we could get some bad binaries. Considering uv builds the entire image, we're opening ourselves up quite a lot, and I would rather not do that.

I use Linux/x86-64 so I don't know anything about multi-platform builds, but the third bullet in "Remediation steps" at this link describes how to get the manifest list hash so we can pin that: https://github.com/ossf/scorecard/blob/40576783fda6698350fcbbeaea760ff827433034/docs/checks.md#pinned-dependencies

Edit: and please put a comment above the `COPY` that describes how to get the sha256 when you have figured out how to do it, otherwise you'll be the one who has to update uv across all ODC repositories forever.

omad: Thanks, I didn't know about `--link`. I'm still not sure exactly what it's for, but the docs say to use it unless you've got symlinks to deal with.

We don't have pinned digests in all our `Dockerfiles` yet, so I was going to just ignore the pinning issue. But, I've managed to make it work.

You can get the digest for a docker image by running `docker buildx imagetools inspect ghcr.io/astral-sh/uv:0.8.15`.

Dependabot will update the tag and digest. (It won't add a digest if it's not already there)

pjonsson: @omad If you're curious, there's a picture for the difference between with and without `--link` about a page down at  https://www.docker.com/blog/image-rebase-and-improved-remote-cache-support-in-new-buildkit/

And I think this version looks fabulous, especially since Dependabot will deal with updating uv in the Docker image from now on!

pjonsson (approved):

PR #840, opened 2025-09-07
stores: check for product earlier

pjonsson:
Without this, the 404 can't be
returned from the dataset-timelines
test.

----
📚 Documentation preview 📚: https://datacube-explorer--840.org.readthedocs.build/en/840/


pjonsson:
Not required, my test was using the wrong URL.

PR #843, opened 2025-09-07
generate: fix pool for pytest-cov

pjonsson:
Fix this now that I saw it since
this stuff is tricky to diagnose
the root cause of once things are
suddenly not shutting down as they
should.

----
📚 Documentation preview 📚: https://datacube-explorer--843.org.readthedocs.build/en/843/


omad (approved):

omad:
Good catch!

PR #844, opened 2025-09-07
generate: autodetect number of jobs

pjonsson:
Detect the number of cores in the machine,
but cap the default value to 16.

Cap only applies to the default value, so
if a user specifies --jobs 48, they will
get 48 workers.

----
📚 Documentation preview 📚: https://datacube-explorer--844.org.readthedocs.build/en/844/

### DIFF
--- a/cubedash/generate.py
+++ b/cubedash/generate.py
@@ -61,6 +61,7 @@ from collections.abc import Generator, Sequence
 from dataclasses import dataclass
 from datetime import timedelta
 from functools import partial
+from os import sched_getaffinity
 from textwrap import dedent
 
 import click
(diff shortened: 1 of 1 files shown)


omad (commented):
I like the idea of setting this automatically, but unfortunately this isn't a suitable way of doing it.

- The number of concurrent processes is limited by the database server, which is very often not on the same machine as the python process.
- CPU Scheduler Affinity is a rarely used mechanism for restricting a process to one or more specific CPU cores, which is different from how much CPU time a process can use. If we did need to know how much processor time we could use in *this* process we'd want to use either https://docs.python.org/3.13/library/os.html#os.process_cpu_count (Python 3.13+ only) or [`dask.system.cpu_count`](https://github.com/dask/dask/blob/main/dask/system.py)

PR #857, opened 2025-09-09
Add type ignores

pjonsson:
These type ignores are for things
we know exist, but cannot convince
the type checker of.

Also add a commit that adds a type
signature, and another commit that
fixes the method chaining order.

----
📚 Documentation preview 📚: https://datacube-explorer--857.org.readthedocs.build/en/857/

### DIFF
--- a/cubedash/index/postgis/_api.py
+++ b/cubedash/index/postgis/_api.py
@@ -77,7 +77,7 @@ class ExplorerIndex(ExplorerAbstractIndex):
         (the ones returned by the Index are cached and so may be shared among callers)
         """
         # why not do md.dataset_fields?
-        return self.index._db.get_dataset_fields(md.definition)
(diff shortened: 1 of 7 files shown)


omad (changes requested):
I think all these `._active_connection()` usages are a more significant problem. 

Most of these functions return a SQLAlchemy Result object that hasn't fetched the results yet, and relies on the connection staying active by, as best I can tell, good luck and best wishes, before the calling function actually returns the result.

The `with` context block looks like it's managing the connection, but is actually doing nothing of the sort.

I'm cleaning up a branch split off my xdist/testing branch that replaces all of these with actual explicit connection handling, and retrieving results while the connection is still in scoppe and open.

omad:
Closing (for now at least).

I'll try to raise a PR today with my connection handling changes which touches all the same lines.

PR #860, opened 2025-09-10
Pytest xdist with parallel postgres docker containers

omad:
WIP

This PR fixes the old support for spinning up a bunch of postgis docker container for testing against in parallel. It ended up broken when the new postgis 1.9 driver was integrated.

It allows running tests on any machine with docker, by running `uv run pytest` and to use pytest-xdist to run the tests significantly faster by splitting them across multiple processes by specifying `--numprocesses auto` (or replace `auto` with a number).

The `cubedash.testutils.database` module provides fixtures to:
- start a new postgis docker container running per test session (1 by default, or more if using `xdist` `--numprocesses`)
- Create a new test ODC database per test module.

`xdist` has a couple of styles of splitting, which both work for me. 

- `--dist loadscope` splits the tests up while running all tests from the same test module on the same test worker.
- `--dist load` ignore modules and splits the tests more evenly across test workers.

The fixtures and usage of `cubedash.testutils.database` are such that either option works. On my 2024 MacBook Pro (12 ARM cores), `loadscope` completes the full test suite in about 5 minutes, and `load` in about 5 minutes. However `load` causes way more load and churn, since duplicate test databases will be setup multiple times for the same test module on different workers.

When running with `xdist`
- Benchmark tests are disabled
- The `--pdb` flag is disabled.
- The fans on your machine will get noisy

----
📚 Documentation preview 📚: https://datacube-explorer--860.org.readthedocs.build/en/860/


.github/workflows/xdist_tests.yml
```diff
@@ -0,0 +1,53 @@
...
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          fetch-depth: 0
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@4959332f0f014c5280e7eac8b70c90cb574c9f9b # v6.6.0
+
+      - name: Install Dependencies
+        run: |
+          uv sync
```
pjonsson, line 43: I think this is missing `--locked --extra=test`.

.github/workflows/xdist_tests.yml
```diff
@@ -0,0 +1,53 @@
...
+
+      - name: Run Tests
+        run: |
+          uv run pytest -v -x integration_tests/ --numprocesses 4 --dist load --durations 20 --junit-xml pytest_junit.xml --cov=--cov-report=xml -r sx
```
pjonsson, line 47: Github's Linux runners have 2 cores, is 4 a good number for these tests?

cubedash/testutils/database.py
```diff
@@ -19,129 +20,252 @@
...
 from datacube.utils import read_documents
 from datacube.utils.documents import InvalidDocException, UnknownMetadataType
+from psycopg2 import sql
 from sqlalchemy import text
 
+logger: structlog.BoundLogger = structlog.get_logger(__name__)
+
+
 GET_DB_FROM_ENV = "get-the-db-from-the-environment-variable"
+POSTGIS_IMAGE = "postgis/postgis:16-3.5"
```
pjonsson, line 30: I'm a bit afraid of having many places to update for the same piece of information because they tend to get out of sync, can this be read from `docker-compose.yaml` instead so things can't go out of sync?

cubedash/testutils/database.py
```diff
@@ -19,129 +20,252 @@
...
+            "-c",
+            "maintenance_work_mem=64MB",
+            # Skip WAL archiving and replication setup
+            "-c",
+            "wal_level=minimal",
+            "-c",
+            "max_wal_senders=0",
+            "-c",
+            "archive_mode=off",
+            # These are redundant with tmpfs but included for completeness
```
pjonsson, line 68: There is no tmpfs support in Docker for macOS, so I assume you get a regular file system there.

cubedash/testutils/database.py
```diff
@@ -19,129 +20,252 @@
...
         )
+
+        log = log.bind(db_url=new_postgresql_url)
+        log.info("Creating postgis extension")
+        conn = psycopg2.connect(new_postgresql_url)
+        # Prevent starting a transaction, since 'create extension' cannot run inside a transaction
+        conn.set_session(autocommit=True)
+        with conn:
+            with conn.cursor() as cur:
+                cur.execute("CREATE EXTENSION postgis")
```
pjonsson, line 185: Isn't this already handled by `create_after_schema()` in `_schema.py`?

cubedash/testutils/database.py
```diff
@@ -19,129 +20,252 @@
...
-def env_name(request) -> str:
+def env_name(request: pytest.FixtureRequest) -> str:
     return request.param
 
 
 @pytest.fixture(scope="module")
-def cfg_env(odc_db, env_name) -> ODCEnvironment:
-    """Provides a :class:`ODCEnvironment` configured with suitable config file paths."""
-    return ODCConfig()[env_name]
+def cfg_env(odc_db, env_name: str) -> ODCEnvironment:
```
pjonsson, line 231: This looks like a bug fix that should go in on develop right away?

Edit: initially thought the `odc_db` part was added, not sure how I misread that.

integration_tests/test_stores.py
```diff
@@ -188,13 +189,13 @@ def test_put_get_summaries(summary_store: SummaryStore) -> None:
...
-            product_name,
-            4321,
-            (datetime(2017, 1, 1), datetime(2017, 4, 1)),
-            [],
-            [],
-            {},
-            datetime.now(),
+            name=product_name,
+            dataset_count=4321,
+            duration=(datetime(2017, 1, 1), datetime(2017, 4, 1)),
```
pjonsson, line 194: Might as well remove the parenthesis in the RHS if you're updating the line.

pyproject.toml
```diff
@@ -192,6 +194,9 @@ select = [
 # The file deliberately doesn't put the import at the top, and we can't avoid global overrides
 "docs/conf.py" = ["E402", "A001", "EXE001"]
 
+# asserts are very much expected in pytest tests
+"integration_tests/*.py" = ["S101"]
```
pjonsson, line 198: Not sure what branch you are on, but #856 that adds the flake8-bandit rules also puts S101 in the ignore list.

pjonsson:
Having 4 CI jobs for testing was my workaround for the tests taking 30 minutes to run, and often failing intermittently by the closed cursor. You have fixed the intermittent failures though, and it would be much better to use one job that runs things in parallel.

Starting up a bunch of database containers seems like a quite heavy-handed way that might consume quite a lot of memory. Would it be difficult to instead have a single database container and run the tests against multiple databases inside that container instead?

cubedash/testutils/database.py
```diff
@@ -160,8 +284,8 @@ def odc_test_db(cfg_env):
 
             dc.close()
 
-            # This actually drops the schema, not the DB
-            pgres_core.drop_db(conn)  # pylint:disable=protected-access
+            # Drops the AGDC schema - badly named function
```
pjonsson, line 287: https://github.com/opendatacube/datacube-core/pull/2174

cubedash/testutils/database.py
```diff
@@ -19,129 +20,252 @@
...
+
+    Yields:
+        Ready to go Datacube Object
 
+    """
+    logger.info(
+        "Initialising ODC Schema for testing", odc_db=odc_db, cfg_db_url=cfg_env.db_url
+    )
     index = index_connect(cfg_env, validate_connection=False)
+    # With permissions causes trouble when run multiple times on the same database
```
pjonsson, line 259: I'm not sure what problem you're seeing, but could you be hitting https://github.com/opendatacube/datacube-core/issues/1959?

cubedash/testutils/database.py
```diff
@@ -19,129 +20,252 @@
...
+
+                # assert not test_database_name.endswith('postgis')
+                # if test_database_name.endswith('postgis'):
+                # import web_pdb; web_pdb.set_trace()
+
+                try:
+                    with conn.cursor() as cur:
+                        log.info("Creating database", db_name=test_database_name)
+                        cur.execute(
+                            sql.SQL("CREATE DATABASE {db_name}").format(
```
pjonsson, line 154: I had a dinner break since my earlier comments and I still love the thought of faster testing, but this feels like a big chunk of dynamic stuff that will be really hard to inspect/reproduce when things go mysteriously wrong in the CI.

We can easily create `$(nproc)` test databases when we spin up the current database container (just add a for loop loop around the create database part of the `create_db.sh` script in commit e6e823bc24077aa7f975c0012cff2c9177cd6899), isn't there some place in `conftest.py` (or similar) where we can patch in the worker number into the database-part of the engine connection string?

.github/workflows/xdist_tests.yml
omad: From https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

image

It may have changed at some stage.

cubedash/testutils/database.py
omad: Agreed.

I'll find somewhere better for this, an environment variable, or a pytest argument, or reading from a config, or something.

cubedash/testutils/database.py
omad: There is `tmpfs` support in Docker Desktop on macOS, it uses a linux VM to host the containers. Most things work about the same as on Linux, but there are difference in networking, like not being able to run the container in the host network namespace.

I get about 5000MB/s with tmpfs, and _only_ 2500MB/s without.

cubedash/testutils/database.py
omad: Probably, I was trying to fix an error that was more of a misconfiguration. I'm doing a lot of commit splitting and this will drop out.

integration_tests/test_stores.py
omad: Mmm, no, they're still required to make it a tuple, if they're removed we'd get a `SyntaxError: positional argument follows keyword argument`.


```python
>>> def mine(hello, foo):
...     print(type(hello))
...     print(type(foo))
...
>>> mine('word', foo=('word', 'word'))
<class 'str'>
<class 'tuple'>
>>> mine('word', foo='word', 'word')
  File "<python-input-11>", line 1
    mine('word', foo='word', 'word')
                                   ^
SyntaxError: positional argument follows keyword argument
```

pyproject.toml
omad: I must have been out of date. :)

omad:
> Having 4 CI jobs for testing was my workaround for the tests taking 30 minutes to run, and often failing intermittently by the closed cursor. You have fixed the intermittent failures though, and it would be much better to use one job that runs things in parallel.
> 


As context, my primary motivation was to run the tests faster locally, not just in GHA. It would be really nice if we can use the same approach everywhere.


> Starting up a bunch of database containers seems like a quite heavy-handed way that might consume quite a lot of memory. Would it be difficult to instead have a single database container and run the tests against multiple databases inside that container instead?

That would be nicer, I tried, but, it was unexpectedly difficult with the way `pytest-xdist` works. The test runners are  isolated and don't communicate with each other, and the `controller` doesn't run the `pytest` collection and so doesn't have the context of whether any fixtures are required or when, so can't be responsible for container startup/shutdown either.

I thought I could solve it by getting the first test worker, named `gw0` to manage the container, but it wouldn't know when to shut it down again, which could either fail any still running tests, or be liable to leave it dangling.

cubedash/testutils/database.py
pjonsson: I read the Docker documentation on tmpfs just a couple of weeks ago, so this is presumably faulty Docker documentation.

.github/workflows/xdist_tests.yml
pjonsson: Oh, that has changed, but a pleasant surprise!

Put `$(nproc)` instead of 4 then, so it will automatically adjust with the number of cores in the runners in the future.

PR #862, opened 2025-09-10
generate: always use Pool

pjonsson:
The imap implementation will be
as lazy as one could expect in
Python 3.15, but until then
it is more eager than one would
expect.

Switch from imap_unordered to
using apply_async instead, but
print the results in a deterministic
order and remove the special
case where a different code path
is taken when running on a single
core.

----
📚 Documentation preview 📚: https://datacube-explorer--862.org.readthedocs.build/en/862/

### DIFF
--- a/cubedash/generate.py
+++ b/cubedash/generate.py
@@ -57,11 +57,12 @@ import collections
 import multiprocessing
 import re
 import sys
-from collections.abc import Generator, Sequence
+from collections.abc import Generator, Mapping, Sequence
 from dataclasses import dataclass
 from datetime import timedelta
 from functools import partial
 from textwrap import dedent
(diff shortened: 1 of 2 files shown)


pjonsson:
This is failing because the Click CLI runner does not flush stderr as it should until Click 8.2.1, so the test output is not appearing as expected.

Unfortunately Explorer cannot use Click 8.2.1 because that changed something else causing multiple test failures, but I believe Click 8.3.0 will fix that issue.

Putting this as a draft for now until Click 8.3.0 is released.

pjonsson:
Still doesn't work with the fixes in Click 8.3.2, but Click 8.3.3 should contain a fix that might unblock this.

PR #869, opened 2025-09-16
Allow running tests outside of, and without docker

omad:
Set `CUBEDASH_BYPASS_DOCKER=True` as an environment variable along with DB access environment variables to allow running the integration tests outside of docker, and without spawning PostGIS containers.

Also document this in the README

----
📚 Documentation preview 📚: https://datacube-explorer--869.org.readthedocs.build/en/869/

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ refer to [deployment instructions](https://datacube-explorer.readthedocs.io/en/l
 
 ### How do I modify the CSS/Javascript?
 
-The CSS is compiled from [Sass](https://sass-lang.com/), and the Javascript is compiled from
+The CSS is compiled from [Sass](https://sass-lang.com/), and the JavaScript is compiled from
 [Typescript](https://www.typescriptlang.org/).
 
 Install [npm](https://www.npmjs.com/get-npm), and then install them both:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -151,7 +151,7 @@ refer to [deployment instructions](https://datacube-explorer.readthedocs.io/en/l
 
 ### How do I modify the CSS/Javascript?
 
-The CSS is compiled from [Sass](https://sass-lang.com/), and the Javascript is compiled from
+The CSS is compiled from [Sass](https://sass-lang.com/), and the JavaScript is compiled from
```
pjonsson, line 154: If you're fixing capitalization, there's a second "javascript" in the header that is also wrongly capitalized.

I think TypeScript in the next line is also mis-capitalized.

README.md
```diff
@@ -179,6 +179,35 @@ Install the test dependencies: `pip install -e .[test]`
...
+# Export environment variables so that the tests know which Database to use
+$ export ODC_DEFAULT_INDEX_DRIVER=postgres
+$ export ODC_POSTGIS_INDEX_DRIVER=postgis
+$ export ODC_DEFAULT_DB_URL=postgresql://localhost/odc-explorer-testing
+$ export ODC_POSTGIS_DB_URL=postgresql://localhost/odc-explorer-testing
+$ export CUBEDASH_BYPASS_DOCKER=True
+
+# Finally, run the tests
+$ uv run pytest integration_tests/
+
```
pjonsson, line 207: Don't think you need this newline.

pjonsson (approved):

README.md
omad: 🤪
I should have known there'd be more, I didn't go looking though, it was just a drive by fix. 

I'm going to go back and fix all the wrong and outdated instructions in the readme soon, I'll catch the other things then too.

PR #870, opened 2025-09-16
Fix timezone sensitive tests

omad:
These tests started failing when I started running them outside the usual docker container, or with a different database configuration.

While it's possible there are bugs with the exact bounds of summary data generation, and not just in the test code, I absolutely do not want to go there this week.

The goal here is to make the tests more useful by being less flaky and move onto actual bugs/features/performance improvements.

----
📚 Documentation preview 📚: https://datacube-explorer--870.org.readthedocs.build/en/870/

### DIFF
--- a/cubedash/testutils/database.py
+++ b/cubedash/testutils/database.py
@@ -147,6 +147,12 @@ def odc_test_db(cfg_env):
     # during testing, and need any performance gains we can get.
 
     with index._db._engine.begin() as conn:  # type: ignore[attr-defined]
+        # Some tests are sensitive to the timezone of the database, and expect it to be UTC
+        quoted_db_name = conn.dialect.identifier_preparer.quote(
+            index._db._engine.url.database  # type: ignore[attr-defined]
+        )
+        conn.execute(text(f"ALTER DATABASE {quoted_db_name} SET timezone TO 'UTC'"))
+
         if index.name == "pg_index":
             for table in [
                 "agdc.dataset_location",
--- a/integration_tests/test_page_loads.py
+++ b/integration_tests/test_page_loads.py
@@ -790,9 +790,11 @@ def test_invalid_product_returns_not_found(client: FlaskClient) -> None:
 
 
 @pytest.mark.parametrize("env_name", ("default",), indirect=True)
-def test_show_summary_cli(clirunner, client: FlaskClient, fix_utc_timezone) -> None:
+def test_show_summary_cli(clirunner, client: FlaskClient) -> None:
     """
     You should be able to view a product with cubedash-view command-line program.
+
+    This test expects the database timezone to be UTC
     """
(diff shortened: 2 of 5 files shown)


integration_tests/test_page_loads.py
```diff
@@ -810,7 +812,9 @@ def test_show_summary_cli(clirunner, client: FlaskClient) -> None:
             f"  to {expected_to.isoformat()} ",
         )
     )
-    assert res.output.startswith(expected_header)
+    result_header = "\n".join(res.output.splitlines()[:5])
+    assert expected_header == result_header
+    # assert res.output.startswith(expected_header)
```
pjonsson: Debug leftovers?

integration_tests/test_stac.py
```diff
@@ -636,6 +636,8 @@ def test_stac_collection(stac_client: FlaskClient):
     """
     Follow the links to the "high_tide_comp_20p" collection and ensure it includes
     all of our tests data.
+
+    This test expects the database timezone to be UTC
```
pjonsson, line 640: Does this test use the new fixture implicitly somehow, or is it a note for people looking at this in the future?

cubedash/testutils/database.py
```diff
@@ -147,6 +147,12 @@ def odc_test_db(cfg_env):
     # during testing, and need any performance gains we can get.
 
     with index._db._engine.begin() as conn:  # type: ignore[attr-defined]
+        # Some tests are sensitive to the timezone of the database, and expect it to be UTC
+        quoted_db_name = conn.dialect.identifier_preparer.quote(
```
pjonsson, line 151: Quoting the database name is not done by the corresponding code in datacube-core.

Since multiple downstream applications seem to need to re-invent the stuff already in datacube-core, can you make a PR on datacube-core that alters the timezone for a database so we can get rid of this functionality in Explorer in the future?

pjonsson (approved):
Remove the leftover debug line before merging.

integration_tests/test_stac.py
omad: These are just notes for the future. 

The fixture changes the timezone for the python process, these tests with a note fail if the database isn't set to UTC when data is loaded into it.

cubedash/testutils/database.py
omad: I might not have bothered with the quoting, but I happened to name my test database `odc-explorer-test`, and Postgres requires that to be inside double quotes.

I'll think about it. It's kind of a weird niche testing thing, that I'm not sure is universally useful. On the other hand, datacube-core does take care of things like roles and permissions.

I'd want to do some more research and investigation into whether:
- the database timezone should always be UTC, or if it might make sense to have in a local timezone in some cases.
- if changing it would break things, or be unexpected otherwise
- it might be better as a warning when running `datacube database init` or `system check`

PR #871, opened 2025-09-17
Speedup STAC collections endpoint

### LINKED ISSUE #802: Extremely Slow STAC Collections Endpoint `/stac/collections`
# Issue
I'm having trouble with extremely slow responses from `/stac/collections` on a Datacube Explorer instance we're running. It's taking about 45s to return a response, and is no faster for subsequent responses. We were also having trouble with it running out of memory until we allowed it to use more.

I think this issue existed before #730 but it did touch the relevant code.

# Diagnosis
The problem appears to be caused by a couple of things.

- Some of our products have extremely complex geometries created by Explorer as part of their all of time summary.
  For example, it looks like this

  Image

- The code that's turning Products+Summaries into STAC Collections is extracting the full summary geometry for every Product from the Database, reprojectinging it, storing it, and then throwing it all away and just returning the Bounding Box, since that's all the STAC Collections specification says to include.

I've also found another inefficiency that occurs if pagination is happening on STAC Collections results, the set of all Products/Summary geometry is downloaded a second time into RAM to just to get the count.

# Proposed fixes
The [collections search query](https://github.com/opendatacube/datacube-explorer/blob/develop/cubedash/index/api.py#L137) is only used to return STAC Collections, which means it never needs to return full geometries, only bounding boxes. Bounding Boxes can be efficiently computed within PostGIS. (Ugh, which may not be available... so might need something else).

The lines at https://github.com/opendatacube/datacube-explorer/blob/develop/cubedash/_stac.py#L965-L967 should be replaced with a call to an efficient SQL `count(*)` query.

@whatnick

omad:
As reported in #802, we were experiencing timeouts or extremely slow responses to the `/stac/collections` endpoint. The implementation was doing a lot of extra data transfer and work, by transferring full geometries from PostGIS into Python before simplifying them down to a bbox.

This moves the processing up into PostGIS, and in my testing gives a speedup between 10x and 121x, and potentially more in different environments. I only tested on my laptop.

Running with local PostGIS server, M4 MacBook Pro.

## pytest-benchmark
Loading `/stac/collections` with the data loaded in `test_stac.py`

image

Small dataset, but about 10x faster.

## Running datacube-explorer locally
Using a copy of our database with the horrendously complex geometries

Run the server:
```
uv run gunicorn -w 3 --threads 2 -k gthread --config python:cubedash.gunicorn_config 'cubedash:create_app()'
```

Run benchmark with [hey](https://github.com/rakyll/hey)

Before this change:
```
# 2 workers, 4 requests
$ hey -c 2 -n 4 -t 30 http://127.0.0.1:8000/stac/collections

Summary:
  Total:        23.5620 secs
  Slowest:      11.9025 secs
  Fastest:      11.6424 secs
  Average:      11.7767 secs
  Requests/sec: 0.1698
```

After this change:
```
# For fast, defaults to 50 workers and 200 requests
$ hey -t 30 http://127.0.0.1:8000/stac/collections

Summary:
  Total:        9.8161 secs
  Slowest:      2.6334 secs
  Fastest:      0.2301 secs
  Average:      1.8218 secs
  Requests/sec: 20.3746
```

121x faster.

I haven't added any request caching discussed in the issue, that can be done separately.

Fixes #802 

----
📚 Documentation preview 📚: https://datacube-explorer--871.org.readthedocs.build/en/871/

### DIFF
--- a/cubedash/_stac.py
+++ b/cubedash/_stac.py
@@ -283,11 +283,7 @@ def as_stac_collection(res: CollectionItem) -> pystac.Collection:
         providers=[],
         extent=Extent(
             pystac.SpatialExtent(
-                bboxes=[
-                    res.footprint_wgs84.bounds
-                    if res.footprint_wgs84
-                    else [-180.0, -90.0, 180.0, 90.0]
-                ]
+                bboxes=[res.bbox if res.bbox else [-180.0, -90.0, 180.0, 90.0]]
             ),
             temporal=pystac.TemporalExtent(
                 intervals=[
--- a/cubedash/_stac.py
+++ b/cubedash/_stac.py
@@ -283,7 +283,10 @@ def as_stac_collection(res: CollectionItem) -> pystac.Collection:
         providers=[],
         extent=Extent(
             pystac.SpatialExtent(
-                bboxes=[res.bbox if res.bbox else [-180.0, -90.0, 180.0, 90.0]]
+                # TODO: Find a nicer way to make the typechecker happier
+                # pystac is too specific in wanting a list[float | int]
+                # odc-geo BoundingBox class is a Sequence[float]
+                bboxes=[list(res.bbox) if res.bbox else [-180.0, -90.0, 180.0, 90.0]]
             ),
             temporal=pystac.TemporalExtent(
                 intervals=[
--- a/cubedash/index/postgis/_api.py
+++ b/cubedash/index/postgis/_api.py
@@ -42,7 +42,7 @@ from sqlalchemy import (
     union_all,
     update,
 )
-from sqlalchemy.dialects.postgresql import TSTZRANGE, insert
+from sqlalchemy.dialects.postgresql import TSTZRANGE, array, insert
 from sqlalchemy.orm import aliased
 from sqlalchemy.sql import ColumnElement
 from sqlalchemy.types import TIMESTAMP
@@ -347,27 +347,6 @@ class ExplorerIndex(ExplorerAbstractIndex):
             )
 
     @override
-    def collection_cols(self) -> Select:
-        product_overview = (
-            select(
-                ProductSpatial.name,
-                TimeOverview.footprint_geometry,
-                ProductSpatial.time_earliest,
-                ProductSpatial.time_latest,
-                TimeOverview.period_type,
-            )
-            .select_from(TimeOverview)
-            .join(ProductSpatial)
-            .cte("product_overview")
-        )
-
-        return (
-            select(ODC_PRODUCT.definition, product_overview)
-            .select_from(product_overview)
-            .join(ODC_PRODUCT, product_overview.c.name == ODC_PRODUCT.name)
-        )
-
-    @override
     def collections_search_query(
         self,
         limit: int,
@@ -376,23 +355,49 @@ class ExplorerIndex(ExplorerAbstractIndex):
         bbox: tuple[float, float, float, float] | None = None,
         time: tuple[datetime, datetime] | None = None,
         q: Sequence[str] | None = None,
-    ) -> Result:
-        collection = self.collection_cols().subquery()
-        query = select(collection).where(collection.c.period_type == "all")
-
+    ) -> list[Row]:
+        # STAC Collections only hold a bounding box in EPSG:4326, no geometry
+        # Calculate the bounding box on the server, it's far more efficient
+        collection_bbox = func.Box2D(
+            func.ST_Transform(TimeOverview.footprint_geometry, 4326)
+        )
+        bbox_array = array(
+            [
+                func.ST_XMin(collection_bbox),
+                func.ST_YMin(collection_bbox),
+                func.ST_XMax(collection_bbox),
+                func.ST_YMax(collection_bbox),
+            ]
+        )
+        bbox_or_null = case(
+            (func.ST_XMin(collection_bbox).is_(None), None), else_=bbox_array
+        )
+        query = (
+            select(
+                ODC_PRODUCT.definition,
+                ProductSpatial.name,
+                bbox_or_null.label("bbox"),
+                ProductSpatial.time_earliest,
+                ProductSpatial.time_latest,
+            )
+            .select_from(TimeOverview)
+            .join(ProductSpatial, ProductSpatial.id == TimeOverview.product_ref)
+            .join(ODC_PRODUCT, ProductSpatial.name == ODC_PRODUCT.name)
+            .where(TimeOverview.period_type == "all")
+        )
         if name:
-            query = query.where(collection.c.name == name)
+            query = query.where(ProductSpatial.name == name)
 
         if bbox:
             query = query.where(
-                collection.c.footprint_geometry.intersects(func.ST_MakeEnvelope(*bbox))
+                TimeOverview.footprint_geometry.intersects(func.ST_MakeEnvelope(*bbox))
             )
 
         if time:
             query = query.where(
                 and_(
-                    default_utc(time[0]) <= default_utc(collection.c.time_latest),
-                    default_utc(collection.c.time_earliest) <= default_utc(time[1]),
+                    default_utc(time[0]) <= default_utc(ProductSpatial.time_latest),
+                    default_utc(ProductSpatial.time_earliest) <= default_utc(time[1]),
                 )
             )
 
@@ -400,7 +405,7 @@ class ExplorerIndex(ExplorerAbstractIndex):
             title = SimpleDocField(
                 name="title",
                 description="product title",
-                alchemy_column=collection.c.definition,
+                alchemy_column=ODC_PRODUCT.definition,
                 indexed=False,
                 offset=("metadata", "title"),
             )
@@ -408,7 +413,7 @@ class ExplorerIndex(ExplorerAbstractIndex):
             description = SimpleDocField(
                 name="description",
                 description="product description",
-                alchemy_column=collection.c.definition,
+                alchemy_column=ODC_PRODUCT.definition,
                 indexed=False,
                 offset=("description"),
             )
@@ -419,7 +424,7 @@ class ExplorerIndex(ExplorerAbstractIndex):
                     [
                         title.alchemy_expression.icontains(value),
                         description.alchemy_expression.icontains(value),
-                        collection.c.name.icontains(value),
+                        ODC_PRODUCT.name.icontains(value),
                     ]
                 )
             query = query.where(or_(*expressions))
--- a/cubedash/index/postgis/_api.py
+++ b/cubedash/index/postgis/_api.py
@@ -356,8 +356,13 @@ class ExplorerIndex(ExplorerAbstractIndex):
         time: tuple[datetime, datetime] | None = None,
         q: Sequence[str] | None = None,
     ) -> list[Row]:
-        # STAC Collections only hold a bounding box in EPSG:4326, no geometry
+        # STAC Collections only hold a bounding box in EPSG:4326, no polygons
         # Calculate the bounding box on the server, it's far more efficient
+
+        # The Cubedash Product (which maps to a STAC Collection) doesn't have
+        # any bounding box or geometry attached, all the geometries are in the
+        # TimeOverview table, grouped by different `period_types`. In this case
+        # we use the `period_type=="all"` to get the one that covers all time.
         collection_bbox = func.Box2D(
             func.ST_Transform(TimeOverview.footprint_geometry, 4326)
         )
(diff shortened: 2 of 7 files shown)


cubedash/_stac.py
```diff
@@ -283,11 +283,10 @@ def as_stac_collection(res: CollectionItem) -> pystac.Collection:
         providers=[],
         extent=Extent(
             pystac.SpatialExtent(
-                bboxes=[
-                    res.footprint_wgs84.bounds
-                    if res.footprint_wgs84
-                    else [-180.0, -90.0, 180.0, 90.0]
-                ]
+                # TODO: Find a nicer way to make the typechecker happier
+                # pystac is too specific in wanting a list[float | int]
```
pjonsson, line 287: It's not going to help right now, but it's possible pystac would accept a PR that changes the type annotation from list to Sequence.

pjonsson:
I started looking at the code but there's a lot of code being changed, and there are still test failures. I will take a look again when the tests are not failing.

omad:
> I started looking at the code but there's a lot of code being changed, and there are still test failures. I will take a look again when the tests are not failing.

It's not too much code... 

But there is a massive amount of duplication between the postgis and postgres index modules unfortunately, but using different sqlalchemy table definition styles which makes all the code that's otherwise identical, slightly different. It's unfortunately been added to my mental tidy up list.

cubedash/_stac.py
omad: Yep, I was thinking that's the best solution too.

cubedash/index/postgis/_api.py
```diff
@@ -376,39 +355,65 @@ def collections_search_query(
...
+                alchemy_column=ODC_PRODUCT.definition,
                 indexed=False,
                 offset=("metadata", "title"),
             )
 
             description = SimpleDocField(
                 name="description",
                 description="product description",
-                alchemy_column=collection.c.definition,
+                alchemy_column=ODC_PRODUCT.definition,
```
pjonsson, line 418: The corresponding postgres change also fixes the type of the `offset=` parameter.

cubedash/index/postgis/_api.py
```diff
@@ -376,39 +355,65 @@ def collections_search_query(
...
         time: tuple[datetime, datetime] | None = None,
         q: Sequence[str] | None = None,
-    ) -> Result:
-        collection = self.collection_cols().subquery()
-        query = select(collection).where(collection.c.period_type == "all")
-
+    ) -> list[Row]:
+        # STAC Collections only hold a bounding box in EPSG:4326, no geometry
+        # Calculate the bounding box on the server, it's far more efficient
+        collection_bbox = func.Box2D(
```
pjonsson, line 366: The RHS takes its data from `TimeOverview`, shouldn't this be called `t_o_bbox` (or `overview_bbox`, but that is longer)?

cubedash/index/postgis/_api.py
```diff
@@ -376,39 +355,65 @@ def collections_search_query(
...
+        query = (
+            select(
+                ODC_PRODUCT.definition,
+                ProductSpatial.name,
+                bbox_or_null.label("bbox"),
+                ProductSpatial.time_earliest,
+                ProductSpatial.time_latest,
+            )
+            .select_from(TimeOverview)
+            .join(ProductSpatial, ProductSpatial.id == TimeOverview.product_ref)
```
pjonsson, line 388: Since you have a real database which triggers performance problems at your disposal: I know the order of joins can matter for performance, are these two joins in the right order or should the order be flipped?

cubedash/index/postgis/_api.py
```diff
@@ -376,39 +355,65 @@ def collections_search_query(
...
-
+    ) -> list[Row]:
+        # STAC Collections only hold a bounding box in EPSG:4326, no geometry
+        # Calculate the bounding box on the server, it's far more efficient
+        collection_bbox = func.Box2D(
+            func.ST_Transform(TimeOverview.footprint_geometry, 4326)
+        )
+        bbox_array = array(
+            [
+                func.ST_XMin(collection_bbox),
```
pjonsson, line 371: Not knowing Postgis, this looks really strange to me, isn't there some function that gets all 4 corners of a 2D box?

cubedash/index/postgis/_api.py
```diff
@@ -376,39 +355,65 @@ def collections_search_query(
...
+                func.ST_YMin(collection_bbox),
+                func.ST_XMax(collection_bbox),
+                func.ST_YMax(collection_bbox),
+            ]
+        )
+        bbox_or_null = case(
```
pjonsson: The RHS uses None, so I think the naming is a bit indirect. But this appears to be a single-use variable that visually looks like it might wrap reasonably well at its use-site, can't the naming issue be avoided by just inlining this variable?

cubedash/index/postgis/_api.py
```diff
@@ -376,39 +355,65 @@ def collections_search_query(
...
+            .where(TimeOverview.period_type == "all")
+        )
         if name:
-            query = query.where(collection.c.name == name)
+            query = query.where(ProductSpatial.name == name)
 
         if bbox:
             query = query.where(
-                collection.c.footprint_geometry.intersects(func.ST_MakeEnvelope(*bbox))
+                TimeOverview.footprint_geometry.intersects(func.ST_MakeEnvelope(*bbox))
```
pjonsson, line 397: There is no test coverage for most of these if-branches, but I'll trust you know what you are doing with these changes.

cubedash/index/postgis/_api.py
```diff
@@ -419,7 +424,7 @@ def collections_search_query(
                     [
                         title.alchemy_expression.icontains(value),
                         description.alchemy_expression.icontains(value),
-                        collection.c.name.icontains(value),
+                        ODC_PRODUCT.name.icontains(value),
```
pjonsson, line 429: Since this is a performance PR: does SQLAlchemy avoid generating extra junk when there is no `q`,  or should there be a:
```python
if q:
    query = query.where(or_(*expressions))
```
instead of line 430?

cubedash/index/postgres/_api.py
```diff
@@ -422,40 +402,71 @@ def collections_search_query(
...
+        query = (
+            select(
+                ODC_PRODUCT.c.definition,
+                PRODUCT.c.name,
+                bbox_or_null.label("bbox"),
+                PRODUCT.c.time_earliest,
+                PRODUCT.c.time_latest,
+            )
+            .select_from(
+                TIME_OVERVIEW.join(
```
pjonsson, line 435: I don't know anything about the table layout, are the joins deliberately inside the `select_from()` here but after it in the corresponding postgis code?

cubedash/index/postgres/_api.py
```diff
@@ -464,14 +475,14 @@ def collections_search_query(
...
-                        collection.c.name.icontains(value),
+                        ODC_PRODUCT.c.name.icontains(value),
                     ]
                 )
             query = query.where(or_(*expressions))
 
         query = query.limit(limit).offset(offset)
 
-        with self.index._active_connection() as conn:
+        with self.engine.begin() as conn:
```
pjonsson: This change isn't present in the postgis index.

integration_tests/test_stac.py
```diff
@@ -707,6 +707,38 @@ def test_stac_collection(stac_client: FlaskClient):
...
+# Tests for STAC API - Features/Collections
+# https://api.stacspec.org/v1.0.0/ogcapi-features/
+# https://github.com/radiantearth/stac-api-spec/tree/release/v1.0.0/ogcapi-features
+#
+# These cover the responses available at /stac/collections/*
+# The spec offers either a cut down version called 'Collections', but Explorer
+# supports the full 'Features' specification.
+
+
+@pytest.mark.parametrize("env_name", ("default",), indirect=True)
```
pjonsson, line 725: Since this is a performance issue, shouldn't this also be enabled for the postgis backend?

cubedash/index/postgis/_api.py
omad: It looked really strange to me too, but best I can tell... no, there isn't, this is what everyone does.

The only other way is passing the WKT strings back to Python and parsing the floats out, which, for 4 floats in a box would be totally fine, but doesn't _feel_ totally fine, so I did it this way.

cubedash/index/postgis/_api.py
omad: The joins have to be in this order due to the relationships of the tables. Also, if it's trivial to switch the order, I'd generally trust the query planner to do the right thing.

```
 Hash Join  (cost=35.24..8906.12 rows=120 width=827) (actual time=0.473..251.316 rows=144 loops=1)
   Hash Cond: ((product.name)::text = (dataset_type.name)::text)
   ->  Hash Join  (cost=12.00..1231.01 rows=138 width=5550) (actual time=0.160..2.085 rows=144 loops=1)
         Hash Cond: (time_overview.product_ref = product.id)
         ->  Index Scan using time_overview_pkey on time_overview  (cost=0.29..1218.93 rows=138 width=5516) (actual time=0.011..1.861 ro
               Index Cond: (period_type = 'all'::overviewperiod)
         ->  Hash  (cost=9.65..9.65 rows=165 width=38) (actual time=0.136..0.136 rows=144 loops=1)
               Buckets: 1024  Batches: 1  Memory Usage: 19kB
               ->  Seq Scan on product  (cost=0.00..9.65 rows=165 width=38) (actual time=0.019..0.104 rows=144 loops=1)
   ->  Hash  (cost=21.44..21.44 rows=144 width=779) (actual time=0.197..0.197 rows=144 loops=1)
         Buckets: 1024  Batches: 1  Memory Usage: 123kB
         ->  Seq Scan on dataset_type  (cost=0.00..21.44 rows=144 width=779) (actual time=0.012..0.125 rows=144 loops=1)
 Planning Time: 0.550 ms
 Execution Time: 251.369 ms
```


```mermaid
erDiagram
    direction LR
    CUBEDASH_TIME_OVERVIEW }o--|| CUBEDASH_PRODUCT : product_id
    "CUBEDASH_PRODUCT" ||--|| "ODC_PRODUCT" : name

```

cubedash/index/postgis/_api.py
omad: Sure, that'll work.

cubedash/index/postgis/_api.py
omad: Yeah, there's a least one bug that I found in these if-branches. The 'time' filter was crashing by trying to apply a python datetime function to a SQLAlchemy Column.

I only added one test in this PR. There's room for more...

In this case, `collection` was a subquery that held the required columns. I've removed the sub-query since it was unnecessary obfuscation, and apply the WHERE clause against the table.

cubedash/index/postgis/_api.py
omad: I think that `SimpleDocField` turns a plain string into a length 1 tuple... but I'm not sure.

This is ugly though, I'll make them match.

cubedash/index/postgis/_api.py
omad: This whole section is already inside an `if q:` on line 404.

cubedash/index/postgres/_api.py
omad: They are deliberate, this is pretty much my first time writing SQLAlchemy code. I couldn't get it working with the exact same structure because the types of the Tables are different. In the legacy postgres section they're plain SQLAlchemy Tables. In the newer postgis section they're SQLAlchemy ORM Tables.

Another bug-bear with the postgis section is that it didn't go all the way in defining joins that SQLAlchemy can apply automatically, and it absolutely should. Doing that would probably make this query cleaner.

cubedash/index/postgres/_api.py
omad: Whoops, this crept in from the PR I haven't raised yet, that makes this change everywhere.

The `._active_connection()` pattern looks to me like crashes/weird behaviour waiting to happen. As far as I can tell it doesn't do any real context/connection management, and just leaves open connections hanging around in threadlocal variables.

integration_tests/test_stac.py
omad: We're not using the PostGIS backend yet, and, there's a LOT of integration tests here that aren't testing against the PostGIS index either. I was afraid I'd fall down another rabbit hole if I started expanding the test coverage.

The table structure and query structure are identical between the two, with the exception of the naming of the String `name` field that is used to connect the cubedash product table with the odc product/agdc dataset_type table.

cubedash/index/postgis/_api.py
omad: Hmmm. I might add some comments.

The Cubedash Product (which maps to a STAC Collection) doesn't have any bounding box or geometry attached, all the geometries are in the TimeOverview table, grouped by different `period_types`. In this case we use the `period_type=="all"` to get the one that covers all of time.

cubedash/index/postgis/_api.py
pjonsson: I fixed type annotations around that (and a bunch of usages) in datacube-core, I'm quite sure you need the tuple. A bit surprised the type checker wasn't complaining about these lines.

cubedash/index/postgis/_api.py
pjonsson: Duh, the danger of doing evening reviews.

cubedash/index/postgis/_api.py
omad: Thanks!

pyright was complaining, but mypy wasn't for some reason.

It's fixed now regardless.

cubedash/index/postgres/_api.py
pjonsson: Yeah, keeping that in a separate PR makes it easier to review.

I fully agree about using the active connection looking like something that should be fixed.

pjonsson (approved):

PR #873, opened 2025-09-17
Reduce schema checking functions

omad:
- Remove the `has_schema()` function from `cubedash/summary/_schema.py`, it was unused.
- Merge the duplicate `schema_initialised()` methods into the parent class, and use the newer SQLAlchemy introspection `identify()` function.

There's much more of these files that should and could be merged, but this was a trivial one to start with. Slow steps.

----
📚 Documentation preview 📚: https://datacube-explorer--873.org.readthedocs.build/en/873/


pjonsson (approved):
I've generalized this function in datacube-core, so we should be able to call that in the next release of datacube.

PR #876, opened 2025-09-19
Update to Click 8.3.1+patch

pjonsson:
This should fix the IO errors
that were happening with 8.2
and 8.3.

----
📚 Documentation preview 📚: https://datacube-explorer--876.org.readthedocs.build/en/876/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ dependencies = [
     "cachetools",
     # Tests fail with "ValueError: I/O operation on closed file" with Click
     # 8.2.0 and later.
-    "click<8.2.0",
+    "click @ git+https://github.com/pjonsson/click.git@one-close-only",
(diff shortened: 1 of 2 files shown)


pjonsson:
@omad this is click 8.3.1 + a patch to prevent the StreamMixer's streams from getting closed by other objects.

This repository is the one that reliably triggers the "ValueError: I/O operation on closed file", so could we merge this so the patch gets some stress testing? The patch only touches the testing parts of click, so unless one is using the CliRunner for testing, this is pretty much a stock Click 8.3.1 as far as I know.

I don't mind if this PR is reverted when it's time to do a release of Explorer, with a bit of luck there will be a Click 8.3.2 out by then that we should be using.

SpacemanPaul (approved):
Now core requires click>8.2, we will have to do this until it is fixed upstream.

PR #880, opened 2025-09-21
model: fix log call

pjonsson:
This looks like something that was meant
to be an f-string.

Logging the object does not really make
sense, so remove both the f-string parts
and the logging of the object.

----
📚 Documentation preview 📚: https://datacube-explorer--880.org.readthedocs.build/en/880/


omad (approved):
So much legacy. 

I've been thinking about and tinkering with a logging and metrics overhaul in explorer

It's unfortunately not a priority for me at the moment. 

But the logs and metrics in production are not very useful.

PR #881, opened 2025-09-21
Fix some type errors

pjonsson:
The Flask-caching that has been configured with a NullCache for the past 4-5 years are masking these type errors from MyPy.

----
📚 Documentation preview 📚: https://datacube-explorer--881.org.readthedocs.build/en/881/

### DIFF
--- a/cubedash/summary/_stores.py
+++ b/cubedash/summary/_stores.py
@@ -24,7 +24,7 @@ from pygeofilter.backends.sqlalchemy.evaluate import (
 from pygeofilter.parsers.cql2_json import parse as parse_cql2_json
 from pygeofilter.parsers.cql2_text import parse as parse_cql2_text
 from shapely.geometry.base import BaseGeometry
-from sqlalchemy import Row, RowMapping, func, select
+from sqlalchemy import RowMapping, func, select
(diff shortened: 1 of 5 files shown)


cubedash/summary/_stores.py
```diff
@@ -1173,7 +1171,19 @@ def search_collections(
         for r in self.e_index.collections_search_query(
             name=name, bbox=bbox, time=time, q=q, limit=limit, offset=offset
         ):
-            yield _row_to_collection(r)
+            # the 'r' at the moment has
+            # ('definition', 'name', 'bbox', 'time_earliest', 'time_latest')
+            yield CollectionItem(
+                name=r.name,
+                time_earliest=r.time_earliest.astimezone(default_timezone)
```
pjonsson, line 1178: I don't know where the `CollectionItem` becomes visible for someone looking at Explorer, but grouping as an Australian time zone is probably a bit counter-intuitive for most of Europe/America.

Explorer has always done that though, so this PR just preserves the behavior.

omad: Collection and Item (CollectionItem) are all part of the exposed STAC API.

Collection is STAC terminology for Product and Item is STAC terminology for Dataset.

omad (approved):

PR #909, opened 2025-10-17
Fix sortby format in next link

### LINKED ISSUE #742: STAC sortby extension does not always sort
I just got this failure in the CI on a PR that only touched lines after the tests. All other tests, both before and after `test_stac_sortby_extension[default]`, passed.
```python
_____________________ test_stac_sortby_extension[default] ______________________

stac_client = <FlaskClient <Flask 'cubedash'>>

    @pytest.mark.parametrize("env_name", ("default",), indirect=True)
    def test_stac_sortby_extension(stac_client: FlaskClient) -> None:
        sortby = [{"field": "properties.datetime", "direction": "asc"}]
        rv: Response = stac_client.post(
            "/stac/search",
            data=json.dumps(
                {
                    "product": "ga_ls8c_ard_3",
                    "time": "2022-01-01T00:00:00/2022-12-31T00:00:00",
                    "limit": OUR_DATASET_LIMIT,
                    "_full": True,
                    "sortby": sortby,
                }
            ),
            headers={"Content-Type": "application/json", "Accept": "application/json"},
        )
        assert rv.status_code == 200
        doc = rv.json
        assert doc.get("features")
        for i in range(1, len(doc["features"])):
            assert (
                doc["features"][i - 1]["properties"]["datetime"]
                < doc["features"][i]["properties"]["datetime"]
            )
    
        sortby = [{"field": "properties.datetime", "direction": "desc"}]
        rv: Response = stac_client.post(
            "/stac/search",
            data=json.dumps(
                {
                    "product": "ga_ls8c_ard_3",
                    "time": "2022-01-01T00:00:00/2022-12-31T00:00:00",
                    "limit": OUR_DATASET_LIMIT,
                    "_full": True,
                    "sortby": sortby,
                }
            ),
            headers={"Content-Type": "application/json", "Accept": "application/json"},
        )
        assert rv.status_code == 200
        doc = rv.json
        for i in range(1, len(doc["features"])):
            assert (
                doc["features"][i - 1]["properties"]["datetime"]
                > doc["features"][i]["properties"]["datetime"]
            )
    
        rv: Response = stac_client.get(
            "/stac/search?collection=ga_ls8c_ard_3&limit=5&sortby=assets"
        )
        assert rv.status_code == 400
    
        rv: Response = stac_client.get(
            "/stac/search?collection=ga_ls8c_ard_3&limit=5&sortby=id,-properties.datetime"
        )
        doc = rv.json
        for i in range(1, len(doc["features"])):
            assert doc["features"][i - 1]["id"] < doc["features"][i]["id"]
    
        # use of property prefixes shouldn't impact result
        rv: Response = stac_client.get(
            "/stac/search?collection=ga_ls8c_ard_3&limit=5&sortby=item.id,-datetime"
        )
        assert rv.json == doc
    
        # ignore undefined field
        rv: Response = stac_client.get(
            "/stac/search?collection=ga_ls8c_ard_3&limit=5&sortby=id,-datetime,foo"
        )
        assert rv.json["features"] == doc["features"]
    
        # sorting across pages
        next_link = _get_next_href(doc)
        next_link = next_link.replace("http://localhost", "")
        rv: Response = stac_client.get(next_link)
        last_item = doc["features"][-1]
        next_item = rv.json["features"][0]
>       assert last_item["id"] < next_item["id"]
E       AssertionError: assert '2a7c420c-dc07-42bf-ae91-ad5da19c76f6' < '0f294559-b2cf-423b-98a8-57a5ed1b2dbd'
```

Ariana-B:
Hopefully resolves #742 
The value of `sortby` was not being represented (or subsequently parsed) correctly in the `next` link of GET requests as it was being passed to `flask.url_for` as a list object. It's likely this was the reason why the test result was inconsistent.
Convert to a string and add corresponding logic to the request arg parser to resolve. 

----
📚 Documentation preview 📚: https://datacube-explorer--909.org.readthedocs.build/en/909/

### DIFF
--- a/cubedash/_stac.py
+++ b/cubedash/_stac.py
@@ -513,7 +513,12 @@ def _sort_arg(arg: str | list[str] | list[dict[str, Any]]) -> list[dict[str, Any
         # default is ascending
         return {"field": val.strip(), "direction": "asc"}
 
-    arg_work = arg.split(",") if isinstance(arg, str) else arg
+    if isinstance(arg, str):
+        arg_work = (
+            json.loads(arg.replace("'", '"')) if arg.startswith("[") else arg.split(",")
+        )
+    else:
+        arg_work = arg
     if len(arg_work):
         if isinstance(arg_work[0], str):
             return [_format(a) for a in arg_work]  # type: ignore[arg-type]
@@ -643,8 +648,8 @@ def _handle_search_request(
             _o=next_offset,
             _full=full_information,
             intersects=intersects,
-            fields=fields,
-            sortby=sortby,
+            fields=str(fields) if fields else None,
+            sortby=str(sortby) if sortby else None,
             # so that it doesn't get named 'filter_lang'
             **{"filter-lang": filter_lang},
             filter=filter_cql,
--- a/cubedash/_stac.py
+++ b/cubedash/_stac.py
@@ -648,7 +648,7 @@ def _handle_search_request(
             _o=next_offset,
             _full=full_information,
             intersects=intersects,
-            fields=str(fields) if fields else None,
+            fields=fields,
             sortby=str(sortby) if sortby else None,
             # so that it doesn't get named 'filter_lang'
             **{"filter-lang": filter_lang},
--- a/cubedash/_stac.py
+++ b/cubedash/_stac.py
@@ -521,7 +521,7 @@ def _sort_arg(arg: str | list[str] | list[dict[str, Any]]) -> list[dict[str, Any
         arg_work = arg
     if len(arg_work):
         if isinstance(arg_work[0], str):
-            return [_format(a) for a in arg_work]  # type: ignore[arg-type]
+            return [_format(a) for a in arg_work]
         if isinstance(arg_work[0], dict):
             for a in arg_work:
                 assert isinstance(a, dict)


cubedash/_stac.py
```diff
@@ -513,7 +513,12 @@ def _format(val: str) -> dict[str, str]:
         # default is ascending
         return {"field": val.strip(), "direction": "asc"}
 
-    arg_work = arg.split(",") if isinstance(arg, str) else arg
+    if isinstance(arg, str):
```
pjonsson, line 516: This part seems to fix a type error, so that is usually good.

(I've noticed your PRs sometimes fail on the type checker, you can run mypy locally on the file before pushing with `uv run mypy --follow-imports=silent cubedash/_stac.py`, and for other ODC repositories the `--follow-imports` can be skipped.)

cubedash/_stac.py
```diff
@@ -644,7 +649,7 @@ def next_page_url(next_offset):
             _full=full_information,
             intersects=intersects,
             fields=fields,
-            sortby=sortby,
+            sortby=str(sortby) if sortby else None,
```
pjonsson, line 652: Was this broken before, or is this change required because of the type fix above?

If it is required because of the type fix: does `str()` do what you want it to do for both the `dict` and the `list` type?

Ariana-B: This change isn't really a type fix, it has to do with how `flask.to_url` treats lists:
```
flask.url_for('.stac_search', sortby=["a", "b"])
> '/stac/search?sortby=a&sorby=b'
```
The problem here was then twofold: `request_args.get("sortby")` would only pick up the first sortby parameter; and the value, originally a dict, would be formatted as a string, which the arg parser would not handle correctly (it assumed that the sortby string would be a comma-separated list).
Casting the `sortby` list to a string is the easiest way to make sure that it appears as a single parameter in the next url. I also remove the assumption about the formatting of the string, and convert it back to a list of dicts with `json.loads` if needed.

cubedash/_stac.py
Ariana-B: My local mypy often gives different results to the one run in the CI, despite being the same version.

omad: That's really odd! And probably worth resolving... :-)

pjonsson (approved):

cubedash/_stac.py
pjonsson: @Ariana-B if `uv run mypy <..>` gives you errors that aren't seen in CI, I think there's something wrong in the mypy configuration. Can you please write down the exact command you used to run mypy, the error you got, and the repository where you got it the next time it happens for you?

PR #923, opened 2025-11-03
Improve check_page_limit behaviour

Ariana-B:
According to the [Item Search STAC spec](https://api.stacspec.org/v1.0.0/item-search/#tag/Item-Search/operation/postItemSearch):
>  If the limit parameter value is greater than advertised limit maximum, the server must return the maximum possible number of items, rather than responding with an error.

Update `check_page_limit` to reflect this, and include handling for `limit < 1` for it to be more comprehensive and consistent. This new behaviour extends to Collection Search and /catalogs/arrivals/items  - api.stacspec.org does not seem to document the expected behaviour for Collection Search but it seems reasonable to assume it would be the same across the board.

----
📚 Documentation preview 📚: https://datacube-explorer--923.org.readthedocs.build/en/923/

### DIFF
--- a/cubedash/_stac.py
+++ b/cubedash/_stac.py
@@ -56,16 +56,15 @@ def get_default_limit() -> int:
     return current_app.config.get("STAC_DEFAULT_PAGE_SIZE", DEFAULT_PAGE_SIZE)
 
 
-def check_page_limit(limit: int) -> None:
+def check_page_limit(limit: int) -> int:
+    if limit < 1:
+        return get_default_limit()
     page_size_limit = current_app.config.get(
         "STAC_PAGE_SIZE_LIMIT", DEFAULT_PAGE_SIZE_LIMIT
     )
     if limit > page_size_limit:
-        abort(
-            400,
-            f"Max page size is {page_size_limit}. "
-            "Use the next links instead of a large limit.",
-        )
+        return page_size_limit
+    return limit
 
 
 def dissoc_in(d: dict, key: str):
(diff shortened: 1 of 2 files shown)


cubedash/_stac.py
```diff
@@ -1370,7 +1363,7 @@ def arrivals_items():
     """
     limit = request.args.get("limit", default=get_default_limit(), type=int)
     offset = request.args.get("_o", default=0, type=int)
-    check_page_limit(limit)
+    limit = check_page_limit(limit)
```
pjonsson: If you wrap `check_page_limit()` around the RHS of the previous line that defines `limit`, it will be impossible for future changes to use the wrong limit value.

Same comment for the other places with this change.

cubedash/_stac.py
```diff
@@ -56,16 +56,15 @@ def get_default_limit() -> int:
...
         "STAC_PAGE_SIZE_LIMIT", DEFAULT_PAGE_SIZE_LIMIT
     )
     if limit > page_size_limit:
-        abort(
-            400,
-            f"Max page size is {page_size_limit}. "
-            "Use the next links instead of a large limit.",
-        )
+        return page_size_limit
+    return limit
```
pjonsson: `return min(limit, page_size_limit)` let's you remove the if-statement.

cubedash/_stac.py
```diff
@@ -56,16 +56,15 @@ def get_default_limit() -> int:
     return current_app.config.get("STAC_DEFAULT_PAGE_SIZE", DEFAULT_PAGE_SIZE)
 
 
-def check_page_limit(limit: int) -> None:
+def check_page_limit(limit: int) -> int:
+    if limit < 1:
+        return get_default_limit()
```
pjonsson, line 61: When looking at this, I get the impression that `get_default_limit()` might return a value that is greater than `page_size_limit`. Should the `page_size_limit` check also be in the body of `get_default_limit()`?

Ariana-B: The page size and page size limit values are constants, so I think it's fine to expect that they will have reasonable values.

pjonsson (approved):

PR #929, opened 2025-11-05
Initial hooks for additional STAC collection fields

mpaget:
We'd like to enable explorer/stac as a provider endpoint for [CEOS connected data assets](https://ceos.org/ourwork/workinggroups/wgiss/access/connected-data-assets/). The CEOS folks have recommended the addition of the following fields into an explorer/stac Collection response:

- Providers
- Summaries
- Keywords

Ref: https://github.com/radiantearth/stac-spec/blob/master/collection-spec/collection-spec.md

This PR provides the initial hooks to support these three fields. Defaults (and hardcoded) to `None` values in this PR.

The content for each of these fields is not directly available in an ODC product def so will likely need to be compiled from ODC datasets, possibly in a similar manner to `summary._stores._find_product_fixed_metadata()` -- this can be a separate PR.

I don't have a full test/integration suite for explorer (yet) so I confess this PR is both a preliminary integration test and an opportunity for the package maintainers (and others) to help inform the approach. I look forward to your reviews and comments!

----
📚 Documentation preview 📚: https://datacube-explorer--929.org.readthedocs.build/en/929/


CONTRIBUTING.rst
```diff
@@ -52,7 +52,7 @@ articles, and such.
 Submit Feedback
 ~~~~~~~~~~~~~~~
 
-The best way to send feedback is to file an issue at https://github.com/opendatacube/datacube-ows/issues .
+The best way to send feedback is to file an issue at https://github.com/opendatacube/datacube-explorer/issues .
```
pjonsson, line 55: Can you please make a separate PR with this change so we can merge that right away?

cubedash/_stac.py
```diff
@@ -266,7 +266,40 @@ def as_stac_item(dataset: DatasetItem) -> pystac.Item:
...
+    return p
+
+
 def as_stac_collection(res: CollectionItem) -> pystac.Collection:
+    """Return a pystac.Collection view of an internal CollectionItem"""
+    # Add Providers, Summaries and Keywords
+    # providers: list[dict[str, Any]] | None -> list[pystac.Provider]
+    # summaries: dict[str, Any] | None -> pystac.Summaries
+    # keywords: list[str] | None
+    providers = None
```
pjonsson: If you change this into:
```python
providers = None if res.providers is None else [_provider_from_dict(xx) for ..]
```
you can avoid the negated if-condition, and the "final" value gets assigned to providers right away.

Same comment for summaries a couple of lines below.

cubedash/_stac.py
```diff
@@ -266,7 +266,40 @@ def as_stac_item(dataset: DatasetItem) -> pystac.Item:
...
+    if res.providers is not None:
+        providers = [_provider_from_dict(xx) for xx in res.providers]
+    summaries = None  # or pystac.Summaries.empty()
```
pjonsson: These new variables all seem like single use within this function, might as well put them in the RHS where they are used instead.

cubedash/_stac.py
```diff
@@ -266,7 +266,40 @@ def as_stac_item(dataset: DatasetItem) -> pystac.Item:
     return item
 
 
+def _provider_from_dict(d: dict) -> pystac.Provider:
+    """A custom parser that extends pystac.Provider.from_dict()"""
+    # pystac.Provider.from_dict() does not validate roles against ProviderRole nor create a list
```
pjonsson: What does "nor create a list" mean? The type signature for `Provider.role` is `list[ProviderRole] | None` according to the pystac documentation.

pjonsson:
This looks like the relatively straightforward plumbing part of the functionality, and it's hard to argue with that since the data has to be passed through in some way and you seem to have chosen a perfectly reasonable way. My experience is that issues where something isn't available when you need it tends to come in the non-plumbing part, and that isn't present in this PR.

For merging it's good if there are some tests, but I don't mind taking a look at the code that provides the functionality even without tests, can you push the second part of the PR?

CONTRIBUTING.rst
mpaget: Moved to #930

cubedash/_stac.py
mpaget: Thanks. Done

cubedash/_stac.py
mpaget: Thanks. Done

cubedash/_stac.py
mpaget: Updated comment to help clarify. My intention was that this is a reasonable place to ensure we're passing on values/types that pystac requires. We need to validate role values anyway (because `pystac.Provider` doesn't) so thought we could add a couple of convenience mappings too .. e.g., we can pass in a single string in `provider["roles"]=string`, rather than knowing/enforcing earlier that this element should be `provider["roles"]=[string]`.

mpaget:
> ... My experience is that issues where something isn't available when you need it tends to come in the non-plumbing part, and that isn't present in this PR.

Agreed. In this case the non-plumbing part will require some extra thinking as some of the data will need to come from ODC datasets and possibly some from env-vars or sensible choices. If this plumbing PR can be merged then we can begin on the next part with some structure in place to test/develop against.

cubedash/_stac.py
```diff
@@ -266,7 +266,31 @@ def as_stac_item(dataset: DatasetItem) -> pystac.Item:
     return item
 
 
+def _provider_from_dict(d: dict) -> pystac.Provider:
```
Ariana-B, line 269: If we're going to add validation logic to this, I think it could stand to be a little more robust:
* We should handle the case of `"name" not in d` more gracefully than `Provider.from_dict` does
* If `roles` has been provided as a string, we should warn that the expected format is a list

Alternatively, some or all of this logic could reside in the method we will add to retrieve these values in the first place. This isn't a public function, so we could ensure that we have a valid provider definition at an earlier stage.

cubedash/_stac.py
```diff
@@ -266,7 +266,31 @@ def as_stac_item(dataset: DatasetItem) -> pystac.Item:
...
+        p.roles = []  # reset
+        if isinstance(roles, str):
+            roles = [roles]
+        for r in roles:
+            if r is None:
+                continue
+            try:
+                p.roles.append(pystac.ProviderRole(r))
+            except ValueError as e:
+                _LOG.warning(f"Could not parse Provider role: {e}")
```
Ariana-B, line 285: This doesn't really add any useful information to the ValueError message; I think it should include the Provider name at the very least, and maybe even the Collection name.

cubedash/_stac.py
```diff
@@ -266,7 +266,31 @@ def as_stac_item(dataset: DatasetItem) -> pystac.Item:
...
+    # pystac.Provider.from_dict() does not validate roles against ProviderRole
+    # or ensure roles are in a list (in case we provide a string instead of a list)
+    p = pystac.Provider.from_dict(d)
+    roles = p.roles  # unvalidated
+    if roles is not None:
+        p.roles = []  # reset
+        if isinstance(roles, str):
+            roles = [roles]
+        for r in roles:
+            if r is None:
```
Ariana-B, line 280: I don't think null values should be quietly ignored; either we should add a warning here or let it be handled like any other invalid provider role.

PR #933, opened 2025-11-06
Fix version number in Docker builds

### LINKED ISSUE #897: Unclean Version Information in Published Docker Images

e.g.

omad:
We were getting an unclean git repo inside of the Docker build which is where setuptools_scm determines the version number.  This happend because our .dockerignore was excluding some directories that  exist in the git repo but aren't required in the docker image.

That's reasonable, but it makes it impossible to determine the version number correctly inside of the docker build.

This change add a docker ARG, and determines the version number in the GitHub Actions workflow before building the image.

Closes #897

----
📚 Documentation preview 📚: https://datacube-explorer--933.org.readthedocs.build/en/933/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,6 +59,9 @@ RUN --mount=type=cache,id=opendatacube-uv-cache,target=/root/.cache \
 COPY --link . /build/
 
 ARG ENVIRONMENT=deployment
+ARG EXPLORER_VERSION=""
+
+ENV SETUPTOOLS_SCM_PRETEND_VERSION=${EXPLORER_VERSION}
 # The deployment image should not have binaries that aid an attacker to get their
 # rootkit in place, and uv downloads over the network. There is no conditional
 # copy in Docker, so truncate the uv binaries to 0 bytes to render them harmless
@@ -66,6 +69,7 @@ ARG ENVIRONMENT=deployment
 RUN --mount=type=cache,id=opendatacube-uv-cache,target=/root/.cache \
     EXTRAS=$( ([ "$ENVIRONMENT" = "deployment" ] && echo "--extra=deployment --no-dev") || \
                  echo "--extra=test") \
+    && export SETUPTOOLS_SCM_PRETEND_VERSION="${EXPLORER_VERSION}" \
     && uv sync --frozen $EXTRAS --no-editable \
     && ([ "$ENVIRONMENT" != "deployment" ] || \
         (chmod 644 /usr/local/bin/uv* && \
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -74,6 +76,9 @@ RUN --mount=type=cache,id=opendatacube-uv-cache,target=/root/.cache \
 
 FROM base
 
+ARG EXPLORER_VERSION=""
```
pjonsson, line 81: This will invalidate the docker cache from this line and onwards.

pjonsson:
I want good version numbers, but I'm skeptical to the approach in this PR. What is missing for setuptools_scm to behave correctly inside the image, or what things would need to be removed from `.dockerignore` for it to work?

omad:
> I want good version numbers, but I'm skeptical to the approach in this PR. What is missing for setuptools_scm to behave correctly inside the image, or what things would need to be removed from `.dockerignore` for it to work?

For setuptools_scm to operate correctly in the build, it needs to have a clean checkout of the tagged/released git repo.

There's a whole bunch of things that are being ignored at the moment that exist in the repo.

This is clearly not quite right, but what would you propose as a better fundamental approach?

I did also consider building a wheel outside of the docker, and then installing that inside.

I'm not at all happy with the current Dockerfile, I really don't like the combined development/deployment approach. If we were to keep it combined, it should be done using `--target` to specify which variant to build, rather than an ARG.

pjonsson:
How about switching version numbering scheme (https://setuptools-scm.readthedocs.io/en/latest/extending/#version-number-construction) to `no-guess-dev` and `local_scheme` to `no-local-version`? The changes in #935 makes the page report v3.1.1 if I tag a commit and build a deployment image, and adds a `.post0.devN` if I add a commit on top of the commit that the tag points to and rebuild the image.

Dockerfile
```diff
@@ -59,13 +59,17 @@ RUN --mount=type=cache,id=opendatacube-uv-cache,target=/root/.cache \
 COPY --link . /build/
 
 ARG ENVIRONMENT=deployment
+ARG EXPLORER_VERSION=""
+
+ENV SETUPTOOLS_SCM_PRETEND_VERSION=${EXPLORER_VERSION}
```
pjonsson, line 64: This should not be required, it's supposed to be handled by line 72.

Makefile
```diff
@@ -70,25 +70,30 @@ clean:  ## Clean all working/temporary files
 
 # DOCKER STUFF
 up: ## Start server using Docker
-	docker compose up --quiet-pull
+	export EXPLORER_VERSION=$$(uvx setuptools-scm) && \
```
pjonsson, line 73: I think we should have something like this earlier in the Makefile:
```Makefile
ifdef NO_UV
VERSION_CMD = git describe --tags | sed 's#\-g#+g#'
else
VERSION_CMD = uvx setuptools-scm
endif
```
and then use `$(VERSION_CMD)` inside the parenthesis in the places that now have `$$(uvx setuptools-scm)`. This would make it possible to build the image even for someone who doesn't have uv installed on their machine (me, as one example).

omad:
Fixed by #1015

pjonsson:
@omad I just lifted the first commit from this into the linked PR, so we have the version numbers, but your improvements on top of that aren't merged.

omad:
Oh, in that case, I'll have another look next week.

Thanks @pjonsson

PR #935, opened 2025-11-07
Switch to no-guess-dev version scheme

### LINKED ISSUE #897: Unclean Version Information in Published Docker Images

pjonsson:
This avoids counting the dirty bit
when computing the version number,
so the partial repository inside
the Docker image doesn't matter.

Fixes #897

----
📚 Documentation preview 📚: https://datacube-explorer--935.org.readthedocs.build/en/935/


omad:
Interesting.. but there's a few big drawbacks with this too.

There's no indication in the version number whether the build directory was clean or dirty. Even when building outside of Docker.

I think that what we actually need to do is either get rid of or greatly shrink the `.dockerignore` file. We're "building" the package inside of docker, and to do that, we need a clean git directory, and working around that is just tying us up in knots.

We manage what gets included in the built docker image in other ways, and .dockerignore is unnecessary. There's small efficiency wins with having it not bloat the build context. For example, with `uv` it makes a lot of sense to exclude `.venv/`, otherwise that will all get sent to the builder.

PR #941, opened 2025-11-15
CI: use anchors for paths

pjonsson:
The path lists are meant to be identical,
so use the new yaml anchors to ensure that.

----
📚 Documentation preview 📚: https://datacube-explorer--941.org.readthedocs.build/en/941/

### DIFF
--- a/.github/workflows/deployment_test.yaml
+++ b/.github/workflows/deployment_test.yaml
@@ -3,15 +3,14 @@ name: Docker Image
 
 on: # yamllint disable-line rule:truthy
   pull_request:
-    paths:
+    paths: &paths
       - '**'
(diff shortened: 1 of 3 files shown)


omad (commented):
I'm not 100% sure the cognitive overhead of YAML anchors is worth it here...

_(30s pause to read it again)_ 

Can't we delete those path lines entirely? Isn't `**` the default?

pjonsson:
I'm not sure about the default, but the CI jobs in this repository are quite slow so I went ahead and added a bunch of paths so the jobs are only triggered when something that they test has changed.

SpacemanPaul (approved):

PR #942, opened 2025-11-17
Don't show hidden product count in main heading

jeremyh:
On the primary header of the primary page of Explorer, users are shown a product count including the number of hidden products:

image

But the point of hidden products was to make them _not_ be very visible, so this seems like a rather loud place to mention them. The current styling/spacing of this subheading is also inconsistent with the rest of the page, so it doesn't present Explorer well as a first impression.

Move the hidden products to a section at the bottom of the page, after the main content, instead of the top.

----
📚 Documentation preview 📚: https://datacube-explorer--942.org.readthedocs.build/en/942/


pjonsson:
Makes sense, if you update the failing tests I'm happy to approve.

jeremyh:
I should have run beforehand -- tests are passing now

SpacemanPaul (approved):

PR #962, opened 2025-11-30
Include psycopg3 in dependencies

pjonsson:
Make it possible to use psycopg3
with the default packages of
Explorer.

Nothing will use this by default,
but this will make it easier for
people to test psycopg3 with
the next release of Explorer.

----
📚 Documentation preview 📚: https://datacube-explorer--962.org.readthedocs.build/en/962/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,6 +52,7 @@ RUN --mount=type=cache,id=opendatacube-uv-cache,target=/root/.cache \
     uv sync --frozen --extra=deployment --no-install-project \
       --no-binary-package fiona \
       --no-binary-package netcdf4 \
+      --no-binary-package psycopg \
       --no-binary-package psycopg2 \
       --no-binary-package rasterio \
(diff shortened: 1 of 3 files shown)


Ariana-B:
Given that support for psycopg3 is still in its experimental/transitional phase, I wonder if it should live in its own optional dependency group like in datacube-core. Installing both psycopg2 and psycopg3 by default strikes me as odd.

pjonsson:
I think the argument that the tests work with psycopg3 but the functionality hasn't had any wider exposure is pretty compelling for installing both; it lets the user switch between using psycopg2/3 by just changing the connection string, everything else remains exactly the same so there's no possibility to have configured anything else wrong.

Ariana-B:
> the tests work with psycopg3 

By that account, should this PR also update the test suite to run with psycopg3?

SpacemanPaul (approved):

PR #996, opened 2026-01-13
CI: drop permissions on workflow

pjonsson:
----
📚 Documentation preview 📚: https://datacube-explorer--996.org.readthedocs.build/en/996/

### DIFF
--- a/.github/workflows/scan.yml
+++ b/.github/workflows/scan.yml
@@ -7,6 +7,8 @@ on:
 env:
   IMAGE_NAME: ghcr.io/opendatacube/explorer
 
+permissions: {}
+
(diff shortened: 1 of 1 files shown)


Ariana-B:
What's the thought behind this change? The current permissions seem reasonable to me and I don't believe the workflow will be able to run with all permissions dropped.

pjonsson:
This fixes one of the complaints at https://securityscorecards.dev/viewer/?uri=github.com/opendatacube/datacube-explorer

The empty braces still gives read permissions for cloning the repository for some reason, and the job itself has write-permissions for security-events. (The scan.yml used by OWS drops them the same way as this PR.)

SpacemanPaul (approved):

Issue #896, opened 2025-10-07
Publish Docker Images to GHCR


omad:
Switch docker image publishing to GHCR.

We can push to both DockerHub and GHCR for a while, the next release, or something like that.

pjonsson:
It's still untested, but a couple of months ago I made a PR for datacube-docker that should solve this problem for that repository: https://github.com/opendatacube/datacube-docker/pull/171

PR #1001, opened 2026-01-19
Update docker workflow

Ariana-B:
Update the docker workflow to push to GHCR instead of DockerHub (as per #896), switch to using `docker/build-push-action` and add attestation to be more in line with other ODC packages (e.g. OWS)

----
📚 Documentation preview 📚: https://datacube-explorer--1001.org.readthedocs.build/en/1001/

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -7,7 +7,6 @@ on:
   push:
     branches:
       - develop
-      - integrate_1.9
     paths:
       - "**"
 
@@ -16,6 +15,7 @@ on:
 
 env:
   IMAGE_NAME: opendatacube/explorer
+  REGISTRY: ghcr.io
 
 permissions: {}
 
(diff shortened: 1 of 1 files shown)


omad (approved):
Thanks Ariana, looks good  to me.

Attestation and GHCR is awesome! As is dropping the third/fourth/whatever party docker-build-with-cache action.

Not for now, just thoughts for someone in the future:
- I was looking at this last year, and...  ended up going in circles and getting nowhere. I really don't love the combination of GitHub Workflow + Actions + Docker for having a maintainable and understandable release.
- I don't love the split/duplication between release/non-release in the same workflow. It's hard to reason about what's happening. Which also means I'm not sure how best to change it. 
- To get really effective caching we need to switch away from the `gha` cache, because it doesn't get shared between branches for security reasons. We'd have to push cache images to GHCR with all the ephemeral layers included.

pjonsson:
> To get really effective caching we need to switch away from the `gha` cache, because it doesn't get shared between branches for security reasons.

I think we could safely set `cache-from` to include the cache from the current branch and the develop branch, it's just that nobody has done it.

PR #1004, opened 2026-01-23
Use ODC db roles

SpacemanPaul:
Use the native ODC database roles (`agdc_user`, etc for postgres, `odc_user`, etc for postgis), with the native ODC `user` type replacing `explorer_viewer`, the native `manage` type replacing `explorer_generator`, and the native `admin` type replacing `explorer_owner`.

Permissions management is now handled internally by `cubedash-gen --init`, replacing the old SQL script directory.  Inital schema creation or migration to new permission framework will require database superuser permissions, subsequent schema updates should only required the native ODC `admin` role.

If the legacy `explorer_*` roles exist, the new roles are granted to them, for cross-compatibility.

This brings explorer into line with core and OWS for db user/permissions management.

Still in draft for:

- further testing with legacy corner cases, 
- updating documentation (including README)
- misc cleanup and refactoring - some more code could be shared between postgres and postgis index drivers.

Some code should really be pushed into core (and/or consolidated within core) and shared with OWS and Explorer - this will be handled in other future PRs.

----
📚 Documentation preview 📚: https://datacube-explorer--1004.org.readthedocs.build/en/1004/


cubedash/index/api.py
```diff
@@ -219,6 +220,15 @@ def schema_compatible_info(
...
     ) -> tuple[str, bool]: ...
 
+    @abstractmethod
+    def create_schema(self) -> bool: ...
+
+    def drop_all(self) -> None:
+        """
+        Drop all explorer-specific tables/schema
+        """
+        self.execute_ddl(DropSchema(CUBEDASH_SCHEMA, cascade=True, if_exists=True))
```
pjonsson, line 230: There is a `drop_schema()` function in `_core.py` in datacube-core, I think that should be used here.

cubedash/index/postgis/_api.py
```diff
@@ -838,9 +839,35 @@ def schema_compatible_info(self, for_writing_operations_too=False):
...
+        with self.engine.connect() as conn:
+            if not _schema.roles_exist(conn, ["odc_user", "odc_manage", "odc_admin"]):
+                raise RuntimeError(
+                    "Default datacube users do not exist.  Please run 'datacube system init'"
+                )
+
+        # Create schema if necessary and ensure it is owned by odc_admin
+        if not self.schema_initialised():
+            self.execute_ddl(
+                text(f"create schema {CUBEDASH_SCHEMA} authorization odc_admin")
```
pjonsson, line 854: The schema creation code must be fairly similar for core, OWS and Explorer. Rather than having three (or more) copies of the same code that we need to maintain, isn't it better to put the code in core and let Explorer/OWS use that code?

cubedash/index/postgis/_api.py
```diff
@@ -838,9 +839,35 @@ def schema_compatible_info(self, for_writing_operations_too=False):
...
+            )
+        else:
+            with self.engine.connect() as conn:
+                owner = conn.execute(
+                    "select pg.catalog.pg_get_userbyid(nspowner) from pg_catalog.pg_namespace "
```
pjonsson: There must be a `text()` missing around this?

cubedash/index/postgis/_schema.py
```diff
@@ -241,18 +245,14 @@ def get_srid_name(conn: Connection, srid: int):
...
-    # Note that, as above, we deliberately don't use the built-in "if not exists"
-    #
-    if (
-        conn.execute(
-            text("select count(*) from pg_extension where extname='postgis';")
-        ).scalar()
-        == 0
-    ):
-        conn.execute(DDL("create extension postgis"))
+    conn.execute(DDL("create extension if not exists postgis"))
```
pjonsson: The postgis index in core already requires the postgis extension, can Explorer run with a postgres index from core, or is this property already guaranteed by core nowadays?

If this code needs to run here, this also feels like something that should be available as a helper function in core so not every core client has to dive right into the low level database stuff for "standard" things.

cubedash/index/postgis/_schema.py
```diff
@@ -327,6 +327,65 @@ def create_after_schema(conn: Connection, epsg_code: int) -> None:
     )
 
 
+def ensure_permissions(conn: Connection) -> None:
```
pjonsson: This function grants permissions, so calling it `grant_permissions` would be a better signal of what it does. (Ensuring permissions could just as well mean throw an exception if permissions are missing.)

The generic grants done by this function also feel like they belong in a function in core, and that function should take the schema name and connection as parameters.

cubedash/summary/_schema.py
```diff
@@ -122,3 +126,68 @@ def refresh_supporting_views(conn, concurrently: bool) -> None:
...
+    }
+    n, o, t = defs[object_type]
+    sql = f"select {n}, {o} from {t} where schemaname = '{CUBEDASH_SCHEMA}' and {n} in {tuple(objects)}"
+    for row in conn.execute(text(sql)):
+        if row[1] != new_owner:
+            transfers.append((row[0], row[1]))
+    return transfers
+
+
+def transfer(
```
pjonsson: Not specific enough, call it `transfer_owner` instead.

Edit: or `change_owner`.

cubedash/index/postgres/_api.py
```diff
@@ -913,9 +914,37 @@ def schema_compatible_info(
...
 
+    @override
+    def create_schema(self) -> bool:
+        # Ensure ODC roles exist (roles were formerly optional but not using them is now deprecated)
+        with self.engine.connect() as conn:
+            if not _schema.roles_exist(
+                conn, ["agdc_user", "agdc_manage", "agdc_admin"]
+            ):
+                raise RuntimeError(
+                    "Default datacube users do not exist.  Please run 'datacube system init'"
```
pjonsson: Is the double space after the period intentional?

cubedash/summary/_schema.py
```diff
@@ -122,3 +126,68 @@ def refresh_supporting_views(conn, concurrently: bool) -> None:
...
+
+def roles_exist(conn: Connection, roles: list[str]) -> bool:
+    return all(
+        conn.execute(text(f"select 1 from pg_roles where rolname = '{role}'")).scalar()
+        for role in roles
+    )
+
+
+@contextlib.contextmanager
+def as_role(conn: Connection, role: str | None) -> Iterator[Connection]:
```
pjonsson: `Generator[Connection]`

cubedash/summary/_schema.py
```diff
@@ -122,3 +126,68 @@ def refresh_supporting_views(conn, concurrently: bool) -> None:
...
+                )
+            )
+    except ProgrammingError:
+        _LOG.warning(
+            f"Cannot transfer ownership of table {obj_name} from {current_owner} to {new_owner}: "
+            f"active user is not a superuser or current owner"
+        )
+
+
+def roles_exist(conn: Connection, roles: list[str]) -> bool:
```
pjonsson, line 206: I think this functionality also belongs in core (and there is a `has_role` in core that does the same thing for a single user).

cubedash/index/api.py
SpacemanPaul: There are two - in `drivers/postgis/_core.py` and `drivers/postgres/_core.py`.   If I use one of them here, that will cause problems if/when we remove that duplication in core in future.

I definitely agree that there is a lot of cross-repo rationalisation to be done around schema creation/management between core, OWS and Explorer, but I don't think it belongs in this PR - it needs to start in core and bubble up to OWS and Explorer.

cubedash/index/postgis/_api.py
SpacemanPaul: See comment above.

cubedash/index/postgis/_schema.py
SpacemanPaul: There's some (presumably ancient) comments about the explorer schema possibly being in a different database to the ODC schema, which would then require this - but I'm not sure how separate databases could possibly work in practice - I can't see any allowance for separate SQLA engines for accessing ODC  and Explorer tables.

I'm inclined to agree with this one.

cubedash/summary/_schema.py
SpacemanPaul: As above, I agree, but the change needs to start in core - not in this PR.

cubedash/index/postgis/_api.py
SpacemanPaul: Three separate typos in the SQL too - like I said corner cases still needed testing!

cubedash/summary/_schema.py
```diff
@@ -122,3 +126,98 @@ def refresh_supporting_views(conn, concurrently: bool) -> None:
...
+    else:
+        # Changing materialized view ownership requires superuser OR:
+        #   new owner, who has create permission on cubedash schema.
+        try:
+            with as_role(conn, new_owner) as attempt_conn:
+                attempt_conn.execute(text(sql))
+            return  # Success
+        except ProgrammingError:
+            _LOG.warning(
+                f"Cannot transfer ownership of table {obj_name} from {current_owner} to {new_owner}: "
```
Ariana-B:
```suggestion
                f"Cannot transfer ownership of materialized view {obj_name} from {current_owner} to {new_owner}: "
```

cubedash/summary/_schema.py
```diff
@@ -122,3 +126,98 @@ def refresh_supporting_views(conn, concurrently: bool) -> None:
...
+        "matviews": "materialized view",
+    }
+    sql = f"alter {objs[object_type]} {CUBEDASH_SCHEMA}.{obj_name} owner to {new_owner}"
+    try:
+        # Attempt as session user (hopefully we're a superuser)
+        conn.execute(text(sql))
+        return  # Success
+    except ProgrammingError:
+        _LOG.warning(
+            "Cannot transfer ownership as session user.  Try appropriate set role."
```
Ariana-B:
```suggestion
            "Cannot transfer ownership as session user. Trying with appropriate role..."
```
And maybe 'info' level would be more appropriate?

cubedash/index/postgis/_schema.py
```diff
@@ -327,6 +327,65 @@ def create_after_schema(conn: Connection, epsg_code: int) -> None:
...
+    conn.execute(
+        text(
+            f"grant all privileges on all tables in schema {CUBEDASH_SCHEMA} to odc_admin"
+        )
+    )
+
+    # Check for legacy explorer roles, and grant ODC roles for cross-compatibility
+    if roles_exist(conn, ["explorer_viewer", "explorer_generator", "explorer_owner"]):
+        conn.execute(text("grant odc_manage to explorer_generator"))
+        conn.execute(text("grant odc_admin to explorer_owner"))
```
Ariana-B, line 356: What about explorer_viewer?

cubedash/index/postgres/_schema.py
```diff
@@ -393,53 +397,118 @@ def check_or_update_odc_schema(conn: Connection):
...
-    # Add optional indexes to AGDC if we have permission.
-    # (otherwise we warn the user that it may be slow, and how to add it themselves)
-    try:  # should both already be handled in core
-        pg_create_index(conn, "ix_dataset_added", ODC_DATASET.fullname, "added desc")
-        pg_create_index(
-            conn,
-            "ix_dataset_type_changed",
-            ODC_DATASET.fullname,
-            "dataset_type_ref, greatest(added, updated, archived) desc",
+            raise
```
Ariana-B: I know it's old code, but the behaviour doesn't seem to align with the comment above.

cubedash/index/postgres/_schema.py
```diff
@@ -393,53 +397,118 @@ def check_or_update_odc_schema(conn: Connection):
...
-
-            No dataset-update triggers are installed on the ODC instance, and Explorer does
-            not have enough permissions to add them itself.
-
-            It's recommended to run `datacube system init` on your ODC instance to install them.
-
-            Then try this again.
-        """
+    # TODO: Explorer has no business updating the ODC schema itself - should at least delegate to core code.
+    with as_role(conn, "agdc_admin") as admin_conn:
```
Ariana-B: `admin_conn` has already been passed to this function when it was called from `update_schema`, so I  don't think we need this

cubedash/index/postgis/_schema.py
```diff
@@ -327,6 +327,65 @@ def create_after_schema(conn: Connection, epsg_code: int) -> None:
     )
 
 
+def grant_permissions(conn: Connection) -> None:
```
Ariana-B, line 330: The only difference between the two implementations is the prefix of the role names, so I wonder if we could make that a parameter and move these it into `summary._schema`? Same goes for `ensure_owners`

cubedash/index/postgis/_api.py
```diff
@@ -838,9 +839,37 @@ def schema_compatible_info(self, for_writing_operations_too=False):
...
+            )
+        else:
+            with self.engine.connect() as conn:
+                owner = conn.execute(
+                    text(
+                        "select pg_catalog.pg_get_userbyid(nspowner) from pg_catalog.pg_namespace "
+                        f"where nspname='{CUBEDASH_SCHEMA}'"
+                    )
+                ).scalar()
+            if owner != "odc_admin":
```
Ariana-B, line 863: Can this be kept under the with statement?

cubedash/index/postgis/_api.py
```diff
@@ -850,8 +879,11 @@ def refresh_stats(self, concurrently: bool) -> None:
 
         This is ideally done once after all needed products have been refreshed.
         """
-        with self.engine.begin() as conn:
-            _schema.refresh_supporting_views(conn, concurrently=concurrently)
+        with (
+            self.engine.begin() as conn,
```
Ariana-B: Should/could this use `connect`, for consistency?

cubedash/index/api.py
```diff
@@ -38,7 +39,7 @@ def execute_query_scalar(self, query) -> Any:
             return conn.execute(query).scalar()
 
     def execute_ddl(self, query) -> int:
```
Ariana-B, line 41: This is no longer being used anywhere we are not already using `self.engine.connect()`, so I'm not sure I see the purpose of it anymore.

cubedash/index/postgis/_schema.py
SpacemanPaul: All explorer roles were previously created with the agdc_user/odc_user role granted, so not necessary.

cubedash/index/postgis/_schema.py
SpacemanPaul: These functions are a bit too high level for `summary._schema` IMO, even if they are very similar.

cubedash/index/postgis/_api.py
SpacemanPaul: I've dropped the with statement instead.

cubedash/index/api.py
SpacemanPaul: It's still used in a few places.  It might become redundant after we've pushed more schema handling utils down to core - we can drop it then.

cubedash/index/postgis/_api.py
SpacemanPaul: I think so?  I'm just double checking behaviour.

cubedash/index/api.py
Ariana-B: It feels redundant when used amidst `with self.engine.connect() as conn: conn.execute(...)` statements, but if you're replacing those with `execute_ddl` then that resolves things (potential future redundancy notwithstanding).

Ariana-B (approved):

PR #1014, opened 2026-02-09
Handle invalid collection geometry transforms

Ariana-B:
Workaround fix for collections with tricky geometries breaking the entire `stac/collections` endpoint due to an error when transforming to 4326. Make the endpoint a little more robust by defining a custom `safe_transform` function that will catch the error and set the offending geometry to null until a more appropriate solution can be determined.

----
📚 Documentation preview 📚: https://datacube-explorer--1014.org.readthedocs.build/en/1014/


cubedash/index/postgis/_api.py
```diff
@@ -436,6 +438,7 @@ def collections_search_query(
         query = query.limit(limit).offset(offset)
 
         with self.index._active_connection() as conn:
+            conn.execute(text(TRANSFORM_SAFE_SQL))
```
pjonsson: I don't think creating the function should be done on every search, creating the function is closer to schema updates.

Taking a step back: can't the crash be avoided by just wrapping `conn.execute()` in a try-except statement that catches `ProgrammingError` or similar?

integration_tests/data/datasets/ga_s1_nrb_iw_hh_hv_0-sample.yaml
```diff
@@ -0,0 +1,914 @@
...
+product:
+  name: ga_s1_nrb_iw_hh_hv_0
+
+location:
+  https://deant-data-public-dev.s3.ap-southeast-2.amazonaws.com/experimental/baseline/antarctica/ga_s1_nrb_iw_hh_hv_0/t096_205505_iw3/2024/12/06/ga_s1a_nrb_0-1-0_T096-205505-IW3_20241206T072807Z_stac-item.json
+
+crs: EPSG:3031
+geometry:
+  type: Polygon
+  coordinates: [[[-2.3767950776811214e+06, 2.229244370468855e+06], [
```
pjonsson, line 16: If I feed this into wktmap.com, I get the attached picture, and that seems like a valid polygon. Why can't this be transformed?

Image

cubedash/index/postgis/_api.py
Ariana-B: > I don't think creating the function should be done on every search, creating the function is closer to schema updates.

Agreed, I’m just being clumsy as I try to figure out how to work with custom functions in sqlalchemy and I will clean it up.

> Taking a step back: can't the crash be avoided by just wrapping `conn.execute()` in a try-except statement that catches `ProgrammingError` or similar?

That was my first thought, but I wasn’t able to find any method of isolating and excluding just the problem rows.

integration_tests/data/datasets/ga_s1_nrb_iw_hh_hv_0-sample.yaml
Ariana-B: In the background, Explorer product geometries have to be converted to 3577 to be stored in the database. This is a problem for Antarctic datasets that cross the 3577 antimeridian, making the resultant geometry unable to be transformed to 4326 for stac. Fixing this core design issue will take a lot of work, so all I’m looking to do in this PR is provide a quick fix so that the stac api doesn’t crash.

cubedash/index/postgres/_api.py
```diff
@@ -418,7 +419,8 @@ def collections_search_query(
         # we use the `period_type=="all"` to get the one that covers all time.
 
         collection_bbox = func.Box2D(
-            func.ST_Transform(TIME_OVERVIEW.c.footprint_geometry, 4326)
+            # TODO: reinstate ST_Transform once possible
+            func.cubedash.transform_safe(TIME_OVERVIEW.c.footprint_geometry, 4326)
```
SpacemanPaul: I think you should be able to use your `SafeTransform` class here.

cubedash/index/postgis/_api.py
SpacemanPaul: Agreed, the function only needs to be created once - ideally as part of the schema create/update process.

SpacemanPaul (changes requested):

cubedash/index/postgres/_api.py
SpacemanPaul: nvm

SpacemanPaul (approved):

PR #1015, opened 2026-02-09
Fix version number in Docker builds

### LINKED ISSUE #897: Unclean Version Information in Published Docker Images

pjonsson:
We were getting an unclean git repo inside of the Docker build which is where setuptools_scm determines the version number. This happend because our .dockerignore was excluding some directories that exist in the git repo but aren't required in the docker image.

That's reasonable, but it makes it impossible to determine the version number correctly inside of the docker build.

This change add a docker ARG, and determines the version number in the GitHub Actions workflow before building the image.

Closes #897

----
📚 Documentation preview 📚: https://datacube-explorer--1015.org.readthedocs.build/en/1015/

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -10,6 +10,10 @@ docker-compose.*.yml
 **/__pycache__
 **/*.py[cod]
 
+#Git/VCS
+.git
+.jj
+
 # CI / distribution / packaging
 **/*.egg-info
 .github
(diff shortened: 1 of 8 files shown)


pjonsson:
This is the first half from #933 that fixes the version numbering. The rest of that PR is a cleanup which is nice, but omad is busy so bring in the part that fixes the version numbering.

SpacemanPaul (approved):

PR #1023, opened 2026-02-11
Ensure safe_transform function created with admin ownership.

SpacemanPaul:
Create safe_transform function as admin.

Would need to ensure as for other objects if a release had gone out, but as 3.1.3 never hit PyPI, this should be fine.

(Also the new Core common ownership functions don't support functions yet.)


SpacemanPaul:
Duplicate of #1024

PR #1062, opened 2026-03-18
stac: use datacube's ds2stac

pjonsson:
Replace the old EO3->STAC conversion
code with a call into datacube instead.

----
📚 Documentation preview 📚: https://datacube-explorer--1062.org.readthedocs.build/en/1062/

### DIFF
--- a/cubedash/_stac.py
+++ b/cubedash/_stac.py
@@ -9,9 +9,9 @@ from typing import Any, TypeAlias
 import flask
 import pystac
 import structlog
+from datacube.metadata import ds2stac
 from datacube.model import Range
 from datacube.utils import DocReader, parse_time
-from eodatasets3 import serialise
 from eodatasets3 import stac as eo3stac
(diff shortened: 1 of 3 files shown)


pjonsson:
This makes the test emit a raw `None` in the STAC output for `nodata` value for some raster bands, so putting this in draft for now.

I think the issue will be fixed by https://github.com/opendatacube/datacube-core/pull/2402.

Ariana-B:
This is a little awkward - I was previously working on this same update but for some reason never pushed up the branch.
Since there are some differences in our implementations, I've made it a draft PR so we can compare and consolidate: https://github.com/opendatacube/datacube-explorer/pull/1065

pjonsson:
Well, I don't know what I'm doing in these areas and the PR isn't working so I don't have any particular fondness of keeping this branch. Can you approve #1061 so we get the simple changes merged to start with, and then we work from there?

PR #1063, opened 2026-03-18
tests: update STAC schemas

pjonsson:
The STAC organization have reorganized,
so download the extensions from
the new place.

Also add some new extensions that
will be required when using stac2ds
from datacube-core in the future.

----
📚 Documentation preview 📚: https://datacube-explorer--1063.org.readthedocs.build/en/1063/

### DIFF
--- a/integration_tests/schemas/update.sh
+++ b/integration_tests/schemas/update.sh
@@ -6,6 +6,11 @@ set -eu
 stac_tag='v1.1.0'
 stac_api_tag='v1.0.0'
 
+declare -A exts=(["eo"]="v1.1.0" \
+  ["projection"]="v2.0.0" \
+  ["raster"]="v1.1.0" \
+  ["view"]="v1.0.0" \
+)
 
 function get() {
     echo "$1"
@@ -21,8 +26,6 @@ get 'https://geojson.org/schema/Feature.json'
 stac_version="${stac_tag#v}"
 stac_api_version="${stac_api_tag#v}"
 subfolder="stac-spec-${stac_version}"
-
-set -x
 
 # Clean before updating.
 rm -rf schemas.stacspec.org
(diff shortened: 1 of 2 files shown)


integration_tests/schemas/update.sh
```diff
@@ -48,6 +58,19 @@ ln -s "../schemas.stacspec.org/${stac_api_version}" "stac-api/"
 cd "stac/${stac_version}/item-spec/json-schema"
 wget https://raw.githubusercontent.com/radiantearth/stac-spec/568a04821935cc92de7b4b05ea6fa9f6bf8a0592/item-spec/json-schema/itemcollection.json
 perl -pi -e 's#"const": "0.9.0"#"const": "1.1.0"#g' itemcollection.json
+cd -
+
+# Download STAC extensions.
+rm -rf stac-extensions.github.io
+for ext in "${exts[@]}"; do
```
Ariana-B: I don't think you need both `exts` and `ext_versions`, you should be able to make do with just `ext_versions`: 
```
for ext in "${!exts_versions[@]}"; do
    ext_version=${exts_versions[$ext]}
    ...
```

Ariana-B (approved):

Ariana-B:
Why not add the other extension schemas in this PR as well?

integration_tests/schemas/update.sh
pjonsson: Yep, that's better!

pjonsson:
> Why not add the other extension schemas in this PR as well?

There's 87 of them according to https://stac-extensions.github.io/, so that's a lot of content in the repository that we're not using and need to keep up with. (I'm fairly sure it's only me who has touched the schemas and update script since the first time I did it in 2023.)

Ariana-B:
> > Why not add the other extension schemas in this PR as well?
> 
> There's 87 of them according to https://stac-extensions.github.io/, so that's a lot of content in the repository that we're not using and need to keep up with. (I'm fairly sure it's only me who has touched the schemas and update script since the first time I did it in 2023.)

I meant the eo, view, and projection extension schemas as per the update script, but I just noticed they’re already in the repo - my bad, not sure how I missed that.

PR #1065, opened 2026-03-19
Switch to new stac conversion logic

Ariana-B:
Replace eodatasets stac logic with datacube-core's `ds2stac` method for stac conversion.

Note that this change reduces the amount of information included in telemetry STAC items since `ds2stac` doesn't currently support non-geospatial products - this should be improved with a future core update.

~Switch from orjson to rapidjson for stac item serialisation since orjson doesn't support nan/inf values, making it incompatible with the raster extension.~ Change split out into a separate PR, but comes into effect here.

----
📚 Documentation preview 📚: https://datacube-explorer--1065.org.readthedocs.build/en/1065/


cubedash/summary/_stores.py
```diff
@@ -522,7 +522,14 @@ def _find_product_fixed_metadata(
             iter(self.index.datasets.search(product=product.name, limit=1))
         ).metadata.fields
 
-        simple_field_types = {"string", "numeric", "double", "integer", "datetime"}
+        simple_field_types = {
```
pjonsson, line 526: I'm guessing this is an unrelated bug fix?

Paul was considering doing another Explorer release after the latest core release, so if you split this out into a separate PR this fix should be possible to get into the release.

(And if you put boolean first in the list it's roughly sorted by "simpleness" of the types.)

cubedash/_stac.py
```diff
@@ -158,111 +153,72 @@ def url_for(*args, **kwargs):
...
+            # TODO: needs a rethink/deeper dive - can we do this without projecting to 4326
+            ds.metadata_doc["grid_spatial"]["projection"]["spatial_reference"] = str(
+                dataset.geometry.crs
+            )
+            ds.metadata_doc["grid_spatial"]["projection"]["valid_data"] = (
+                dataset.geom_geojson
+            )
+        item = ds2stac(
+            dataset=ds,
+            base_url=url_for("pages.default_redirect"),
```
pjonsson, line 227: I think you want to hoist the url_for() calls to the top of the function so their result is shared between the two branches like I did in my attempt at this.

cubedash/_stac.py
```diff
@@ -158,111 +153,72 @@ def url_for(*args, **kwargs):
...
-
-    item = eo3stac.to_pystac_item(
-        dataset=dataset_doc,
-        stac_item_destination_url=url_for(
-            ".item", collection=dataset.product_name, dataset_id=dataset.dataset_id
-        ),
-        odc_dataset_metadata_url=url_for("dataset.raw_doc", id_=dataset.dataset_id),
-        explorer_base_url=url_for("pages.default_redirect"),
-    )
+        # add canonical ref pointing to the JSON file on s3
```
pjonsson, line 232: I realize the comment predates your change, but is there anything S3-specific about the link, or would ending the comment with ".. JSON file." without the "on s3" be more truthful?

cubedash/_stac.py
```diff
@@ -158,111 +153,72 @@ def url_for(*args, **kwargs):
...
+        if ds.uri:
+            media_type = "application/json" if ds.uri.endswith("json") else "text/yaml"
+            item.links.append(
```
pjonsson, line 235: Possibly scratch my previous remark about the comment, should there even be a link added here if the uri is something that isn't remotely accessible, like `file://something`?

cubedash/_stac.py
```diff
@@ -158,111 +153,72 @@ def url_for(*args, **kwargs):
...
+    if ds is None or ds._gs is None:
+        # Since we'd have to cobble together a metadata_doc anyway,
+        # creating a pystac.Item directly is probably easier
+        item = pystac.Item(
+            id=str(dataset.dataset_id),
+            geometry=dataset.geom_geojson,
+            bbox=dataset.bbox,
+            properties={
+                "created": utc(dataset.creation_time),
+                "cubedash:region_code": dataset.region_code,
```
pjonsson: I think `cubedash:region_code` is added for both branches near the end of the function.

cubedash/_stac.py
```diff
@@ -158,111 +153,72 @@ def url_for(*args, **kwargs):
...
-                {name: _band_to_measurement(b) for name, b in ds.measurements.items()}
-                if ds is not None
-                else {}
+    if ds is None or ds._gs is None:
+        # Since we'd have to cobble together a metadata_doc anyway,
+        # creating a pystac.Item directly is probably easier
+        item = pystac.Item(
+            id=str(dataset.dataset_id),
+            geometry=dataset.geom_geojson,
+            bbox=dataset.bbox,
```
pjonsson: I've fixed the type annotation for `DatasetItem.bbox` in #1066.

cubedash/_stac.py
```diff
@@ -158,111 +153,72 @@ def url_for(*args, **kwargs):
...
+            Link(
+                rel="collection",
+                target=url_for(".collection", collection=dataset.product_name),
+            )
+        )
+    else:
+        if not ds.is_eo3:
+            # TODO: needs a rethink/deeper dive - can we do this without projecting to 4326
+            ds.metadata_doc["grid_spatial"]["projection"]["spatial_reference"] = str(
+                dataset.geometry.crs
```
pjonsson, line 220: I don't know anything about eo1, but the type error from MyPy on this line seems like a real problem.

pjonsson:
Your PR actually removes lines so I think your code is better (and unlike me you know something about eo1/eo3) so we should go with this PR. I don't know if the tests passed when you were working on this, but you should get a different failure if you run this on top of latest develop compared to what the CI shows here.

cubedash/_stac.py
Ariana-B: I approved it but the change doesn't resolve the mypy error - pystac.Item expects the bbox to be a `list[float]`

cubedash/summary/_stores.py
Ariana-B: Yeah, if I notice small bug fixes I usually just go ahead and include them even if they are unrelated since it's easier than saving them for future PRs. ~I'll split this one out so it's not held up by the rest of this PR.~  Actually, this only potentially affects the `fixed_metadata` field of the product summary (I'm not sure any products are even currently using boolean fields) so there's no real hurry to get it in.

pjonsson: No, don't save them, make the PR right away! It takes 2 minutes to stash the current changes and make a new branch with the fix, and the one-liner fix we're discussing takes 30 seconds to review and merge for someone else, so your fix will most often be merged before you have even made the PR for what you were initially working on.

Here's a link to a post called "On commit messages", but there is a bit about how to (not) organize commits as well in it, it's well worth a read: https://who-t.blogspot.com/2009/12/on-commit-messages.html

cubedash/_stac.py
pjonsson: Oh, you're right. I just fixed the same error for core in https://github.com/opendatacube/datacube-core/pull/2406.

cubedash/summary/_stores.py
Ariana-B: Ok.

pjonsson:
Is the switch from orjson to rapidjson coupled with using ds2stac, or could that migration go into a separate PR that we merge first? (I will probably regret saying this, but since I haven't attempted to switch, it doesn't feel that difficult to review the code changes for the orjson=>rapidjson switch.)

Ariana-B:
> Is the switch from orjson to rapidjson coupled with using ds2stac, or could that migration go into a separate PR that we merge first? (I will probably regret saying this, but since I haven't attempted to switch, it doesn't feel that difficult to review the code changes for the orjson=>rapidjson switch.)

Yes it is coupled with the use of `ds2stac` - more specifically, its implementation of the raster extension. As per the PR description, orjson can cause stac documents implementing the raster extension to be invalid since it doesn't support nan/inf values.

pjonsson:
> Yes it is coupled with the use of `ds2stac` - more specifically, its implementation of the raster extension. As per the PR description, orjson can cause stac documents implementing the raster extension to be invalid since it doesn't support nan/inf values.

Sorry, I meant the other way around. What would break if we just merge the changes that switch from orjson to rapidjson?

Ariana-B:
> > Yes it is coupled with the use of `ds2stac` - more specifically, its implementation of the raster extension. As per the PR description, orjson can cause stac documents implementing the raster extension to be invalid since it doesn't support nan/inf values.
> 
> Sorry, I meant the other way around. What would break if we just merge the changes that switch from orjson to rapidjson?

I don't imagine anything would, but the change doesn't really feel justified/necessary just by itself.
If it's for the sake of sanity-checking that the switch isn't causing any changes to the outputs, I would think that the tests passing with no changes to the expected json outputs other than the addition of the raster information should be proof enough that everything is aligned.

pjonsson:
> I don't imagine anything would, but the change doesn't really feel justified/necessary just by itself. If it's for the sake of sanity-checking that the switch isn't causing any changes to the outputs, I would think that the tests passing with no changes to the expected json outputs other than the addition of the raster information should be proof enough that everything is aligned.

I realize you know all the details since you've implemented this, but my mind isn't as sharp as it once was and career progression amplifies that deficiency by ensuring there's a steady stream of interruptions. I'm trying to find some way to get this into bite-size changes that stand on their own to reduce the risk of missing something.

Ariana-B:
> I realize you know all the details since you've implemented this, but my mind isn't as sharp as it once was and career progression amplifies that deficiency by ensuring there's a steady stream of interruptions. I'm trying to find some way to get this into bite-size changes that stand on their own to reduce the risk of missing something.

Sure, I can split it out.

pyproject.toml
```diff
@@ -15,17 +15,11 @@ dependencies = [
     "flask",
     "flask-cors",
     "flask-themer>=1.4.3",
-    "geoalchemy2>=0.8",
```
pjonsson, line 14: The packages that are imported under `cubedash/` need to be somewhere in the dependencies, otherwise Explorer will break if datacube removes one of these dependencies in some future release.

integration_tests/test_stac.py
```diff
@@ -242,8 +242,23 @@ def _iter_items_across_pages(client: FlaskClient, url: str | None) -> Generator[
...
 def assert_stac_extensions(doc: dict) -> None:
-    for extension_name in doc.get("stac_extensions", ()):
-        get_extension(extension_name).validate(doc)
+    # TODO: remove overwriting of stac_extensions once values have been fixed
+    if extensions := doc.get("stac_extensions"):
+        new_extensions = []
+        for extension_name in extensions:
+            if "/sar/" in extension_name:
+                new_extensions.append(
+                    "https://stac-extensions.github.io/sar/v1.3.0/schema.json"
```
pjonsson, line 251: The extension URI is available as `SCHEMA_URI` from the pystac extensions so you don't need to hardcode the string, here's a line from the tests in datacube:
```python
from pystac.extensions.raster import SCHEMA_URI as RASTER_SCHEMA_URI
```
If you can't use the versions provided by pystac I think you can still get most of the URL from there with something like `new_extensions.append(extension_name.replace("/v1.x.y/", "/v1.3.0/")`.

cubedash/_stac.py
```diff
@@ -158,111 +153,97 @@ def url_for(*args, **kwargs):
...
-                else {}
-            ),
-            accessories=(
-                _accessories_from_eo1(ds.metadata_doc) if ds is not None else {}
-            ),
-            # TODO: Fill in lineage. The datacube API only gives us full datasets, which is
-            #       expensive. We only need a list of IDs here.
-            lineage={},
+        if not ds.is_eo3 and dataset.geometry is not None:
+            # TODO: needs a rethink/deeper dive - can we do this without projecting to 4326
```
SpacemanPaul, line 218: It does seem weird to be reprojecting here.  Shouldn't ds2stac handle that?

SpacemanPaul (approved):
Looks great.  Nice cleanup.

cubedash/_stac.py
Ariana-B: ds2stac does already handle reprojection to 4326, but there's something funky going on with some of the geometries that I haven't been able to figure out. I believe it may have to do with a documented issue around shapes becoming invalid on reprojection: https://github.com/opendatacube/datacube-explorer/blob/develop/cubedash/summary/_stores.py#L1690-L1695

I could add the validity check and buffer workaround to the ds2stac logic, but given it seems to be an eo- and explorer-specific issue, I wonder if it's more appropriate to keep the workaround here.

pjonsson: Is the reprojection something that was introduced by this PR, or is the reprojection always being done but it happens in some other code?

Ariana-B: The current implementation with eo-datasets uses `dataset.geometry.geom` for eo1 datasets, so ultimately I haven't really changed the logic.

pjonsson: I interpret that as the reprojection part we're seeing in this PR is just done elsewhere before, so the state is no worse than before this PR.

I would like to get this into the release, so if you can re-instate the direct dependencies of Explorer (you might be able to just drop that commit) and get this on top of latest develop I think we can merge this unless Paul has objections.

PR #1095, opened 2026-04-08
Use rapidjson for STAC serialisation

Ariana-B:
Split out from #1065

NaN and Inf values are acceptable in STAC documents but orjson will serialise them to null so switch to using rapidjson instead.

----
📚 Documentation preview 📚: https://datacube-explorer--1095.org.readthedocs.build/en/1095/


cubedash/_utils.py
```diff
@@ -500,7 +500,11 @@ def as_json(
 
     response = flask.Response(
         dumps(
-            o, option=OPT_INDENT_2 if prefer_formatted else 0, default=_json_fallback
+            o,
```
pjonsson, line 503: The return type of rapidjson's `dumps` is str whereas orjson returns bytes, does that matter here?

pyproject.toml
```diff
@@ -26,6 +26,7 @@ dependencies = [
     "pygeofilter>=0.2.2",
     "pyproj",
     "pystac",
+    "python-rapidjson",
```
pjonsson, line 26: There are only 2 other calls to orjson, can we replace those as well so we don't have to have two high-performance JSON libraries?

(I think replacing the `dumps(..)` in logs with `dumps(..).encode()` and the `dumps(..).decode("utf-8")` in _filters.py with just `dumps(..)` should give the same wire-format as we have with orjson.)

cubedash/_utils.py
Ariana-B: The response body can be provided as bytes, a string, or an iterable of bytes or strings.

cubedash/logs.py
```diff
@@ -35,7 +33,13 @@ def init_logging(
...
     # own 'default' property that overrides our own.
     def lenient_json_dump(obj, *args, **kwargs):
-        return dumps(obj, option=OPT_SORT_KEYS, default=lenient_json_fallback)
+        return dumps(
+            obj,
+            sort_keys=True,
+            datetime_mode=DM_ISO8601,
+            uuid_mode=UM_CANONICAL,
+            default=lenient_json_fallback,
+        )
```
pjonsson: Can't we tack on an `.encode()` here and leave `__call__` below as it was to get this PR to work?

(I would happily get rid of structlog, but that's for another PR.)

Ariana-B: Even when I fix the structlog error by configuring it with `WriteLoggerFactory` instead of `BytesLoggerFactory`, I get a TypeError because of `click.testing.BytesIOCopy` - so structlog isn't necessarily the issue here.

pjonsson: Oh, I was digging around in that code for fixing what turned out to be the multiple things closing the same buffer. I know there is one fix on Click in the 8.3.3 milestone that will improve things with testing, but I doubt that fix will help with this.

I don't know how much spare time you have, but if you want to go fix Click, it will probably save you some work if you do that after 8.3.3 is released.

pjonsson (approved):

PR #1096, opened 2026-04-08
Update to Click 8.3.2

pjonsson:
This release contains the fix
to only close things once.

Also add a commit that uses
the datacube optional for
performance to ensure
datacube gets the performance
packages it needs.

----
📚 Documentation preview 📚: https://datacube-explorer--1096.org.readthedocs.build/en/1096/

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -119,23 +119,10 @@ jobs:
       - name: Install uv
         uses: astral-sh/setup-uv@5a095e7a2014a4212f075830d4f7277575a9d098 # v7.3.1
 
-      # We need a patched version of click to run tests, but direct dependencies can't
-      # be uploaded to PyPI.
-      #
-      # Therefore we removed the direct click dependency, and simply depend on click.
-      # (and use $SETUPTOOLS_SCM_PRETEND_VERSION to force a clean version number.)
-      #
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -119,23 +119,10 @@ jobs:
...
-      # This step (and the export of SETUPTOOLS_SCM_PRETEND_VERSION below) should be
-      # removed once a click release fixes the CliRunner issue that breaks tests.
-      - name: Remove patched click dependency
-        run: |
-          perl -pi -e 's# \@ git\+https://github\.com/pjonsson/click\.git\@one-close-only##' pyproject.toml
-
       - name: Build Packages
         run: |
-          export SETUPTOOLS_SCM_PRETEND_VERSION=`git describe --abbrev=0 --tags`
+          SETUPTOOLS_SCM_PRETEND_VERSION=$(git describe --abbrev=0 --tags)
```
pjonsson, line 124: OWS 1.9.11 didn't go to PyPi, so keep this part of the workaround to ensure releases get the intended version number and can be uploaded to PyPi.

(Split into using a separate line for exporting to ensure a faulty return value from git is detected.)

SpacemanPaul (approved):

Issue #1100, opened 2026-04-10
Minimal datacube-explorer package for `cubedash-gen`


pjonsson:
Environments/images that run `datacube dataset add ..` should run `datacube-ows-update` and `cubedash-gen --all` after adding the datasets. Those environments need database access through SQLAlchemy/Psycopg, but since they don't run the web server it's only a drawback to have packages like Flask and Gunicorn installed (larger image, more packages with vulnerabilities, more dependencies that can conflict).

It would be nice to move dependencies like Flask in the datacube-explorer package into a `webserver` optional that would be what you install for getting the web server.

Sibling ticket for OWS: https://github.com/opendatacube/datacube-ows/issues/1549

PR #1101, opened 2026-04-10
pyproject: add webserver optional

pjonsson:
Put various packages that shouldn't
be required for running cubedash-gen
into a webserver optional that is
included in the deployment optional.

There still needs to be some decoupling
in the code for being able to run
cubedash-gen completely standalone,
but that can happen in separate
smaller PRs.

Refs #1100

----
📚 Documentation preview 📚: https://datacube-explorer--1101.org.readthedocs.build/en/1101/


pyproject.toml
```diff
@@ -90,6 +82,17 @@ test = [
...
+webserver = [
+    "fiona>=1.10.0",
+    "flask",
+    "flask-caching",
+    "flask-cors",
+    "flask-themer>=1.4.3",
+    "jinja2",
+    "markupsafe",
+    "pyproj",
+    "pystac",
```
Ariana-B, line 95: `fiona` and `pyproj` are both required for refreshing product extents as part of `cubedash-gen`, and don't seem to be webserver-related to begin with.

pjonsson: You're right! I've moved them back.

SpacemanPaul (approved):

PR #1102, opened 2026-04-10
summary: decouple default values

pjonsson:
Put default values into a separate
module so they can be imported
without requiring many dependencies.

Refs #1100

----
📚 Documentation preview 📚: https://datacube-explorer--1102.org.readthedocs.build/en/1102/

### DIFF
--- a/cubedash/_filters.py
+++ b/cubedash/_filters.py
@@ -5,7 +5,6 @@ Common global filters for templates.
 import calendar
 from collections.abc import Iterable, Mapping
 from datetime import datetime, timezone
-from zoneinfo import ZoneInfo
 
 import flask
 import structlog
@@ -16,7 +15,8 @@ from markupsafe import Markup
 from rapidjson import DM_ISO8601, UM_CANONICAL, dumps
 from shapely.geometry import MultiPolygon
 
-from . import _model, _utils
+from cubedash.summary.defaults import DEFAULT_GROUPING_TIMEZONE
+
 from . import _utils as utils
 
 # How far to step the number when the user hits up/down.
@@ -47,7 +47,7 @@ def _get_metadata_center_time(dataset):
 
 @bp.app_template_filter("localised_metadata_center_time")
 def _get_localised_metadata_center_time(date):
-    return date.astimezone(ZoneInfo(_model.DEFAULT_GROUPING_TIMEZONE))
+    return date.astimezone(DEFAULT_GROUPING_TIMEZONE)
 
 
 @bp.app_template_filter("printable_dataset")
@@ -221,7 +221,7 @@ def _max_val(ls):
 
 @bp.app_template_filter("product_license_link")
 def _product_license(product: Product):
-    license_ = _utils.product_license(product)
+    license_ = utils.product_license(product)
 
     if license_ is None:
         return "-"
@@ -273,7 +273,7 @@ def _field_step(field: Field):
 
 @bp.app_template_filter("only_alnum")
 def only_alphanumeric(s):
-    return _utils.only_alphanumeric(s)
+    return utils.only_alphanumeric(s)
 
 
 @bp.app_template_filter("timesince")
--- a/cubedash/generate.py
+++ b/cubedash/generate.py
@@ -78,7 +78,8 @@ from datacube.ui.click import environment_option, pass_config
 from typing_extensions import override
 
 from cubedash.logs import init_logging
-from cubedash.summary import GenerateResult, SummaryStore, UnsupportedWKTProductCRSError
+from cubedash.summary._errors import UnsupportedWKTProductCRSError
+from cubedash.summary._stores import GenerateResult, SummaryStore
 from cubedash.summary.defaults import DEFAULT_EPSG, DEFAULT_TIMEZONE
 
 TYPE_CHECKING = False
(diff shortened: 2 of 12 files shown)


cubedash/generate.py
```diff
@@ -76,14 +78,13 @@
...
-from cubedash.summary import (
-    GenerateResult,
-    SummaryStore,
-    TimePeriodOverview,
-    UnsupportedWKTProductCRSError,
-)
-from cubedash.summary._stores import DEFAULT_EPSG
-from cubedash.summary._summarise import DEFAULT_TIMEZONE
+from cubedash.summary._errors import UnsupportedWKTProductCRSError
+from cubedash.summary._stores import GenerateResult, SummaryStore
```
Ariana-B, line 82: Why separate these imports instead of importing them from `cubedash.summary` as before?
To be fair there are a number of places throughout the code that do this but I don't see why they too shouldn't be consolidated.

cubedash/_filters.py
```diff
@@ -16,7 +15,9 @@
 from markupsafe import Markup
 from shapely.geometry import MultiPolygon
 
-from . import _model, _utils
+from cubedash.summary.defaults import DEFAULT_GROUPING_TIMEZONE
+
+from . import _utils
```
Ariana-B: Can we remove this weird double import?

cubedash/generate.py
```diff
@@ -76,14 +78,13 @@
...
+from cubedash.summary.defaults import DEFAULT_EPSG, DEFAULT_TIMEZONE
+
+TYPE_CHECKING = False
+if TYPE_CHECKING:
+    from cubedash.summary._model import TimePeriodOverview
```
Ariana-B, line 87: Same as above, I don't understand why this import is now separate, nor why there would be a circular import.

pjonsson: When we had more type errors in the code, running `uv run mypy cubedash/somefile.py` resulted in "238 errors in 19 files", and running `uv run mypy cubedash/someotherfile.py` also resulted in "238 errors in 19 files", so basically no matter which file you type checked it would import everything else in Explorer.

Everything needing everything else indicates some kind of structural problem, either things that belong together have been separated into different modules, or things don't belong together but for some reason they are messing with internals of other things (which they probably shouldn't be doing). Both scenarios make the code harder to understand and increase the risk of breaking things with seemingly innocent changes, and that is bad for everyone. I don't know the code base for Explorer well enough to diagnose the root cause, but even if I did, fixing such an issue takes time because there's usually a lot of untangling involved in the process.

This particular PR is about me wanting to run `cubedash-gen` in a container without Flask, Gunicorn and a ton of other packages. Importing `cubedash.summary` imports 4 modules and all their dependencies (which is effectively the rest of Explorer), whereas importing `cubedash.summary._errors` imports a single module without any dependencies.

On a practical level, the code that imports modules in Python is also slow, so it's fairly easy to noticeably reduce the startup time for a CLI by importing fewer things.

This PR doesn't make `cubedash-gen` run without most dependencies, but I resisted the urge to just rewrite everything to keep things manageable with the review. The PR does cut away some parts of the import graph, so it is a step forward even if it is a small one.

cubedash/_filters.py
pjonsson: I agree that these things should be fixed, but I fond a dozen of these kind of strange things just while working on this PR so there must be hundreds of them in the code base. Review time is superlinear to PR size, so it's better to fix in separate PRs later.

cubedash/generate.py
pjonsson: In a not very surprising turn of events it turns out I don't know how Python imports work; there's an implicit import of every `__init__.py` in  the "path" of a package which I'm guessing was what you were hinting at with your question.

I have things to say to the language designers, but since this is how Python works we'll have to deal with it somehow. I added a commit that moves `create_app` into `startup_utils` which mirrors how OWS is structured; the change is small and it's hard to argue that the user probably wants flask loaded when they run `cubedash-gen`. I'm not sure how the `__init__.py` in `summary` should be dealt with, so leaving that for later.

Attaching the dependency graph from `pydeps --reverse` when run on `generate.py` on the develop branch and on this branch (I think the boxes are collapsing multiple things).

cubedash_generate_develop
cubedash_generate_decoupled

cubedash/_filters.py
Ariana-B: I'm not suggesting you should fix all instances of these kinds of patterns in this PR, but given you're already touching this line as part of the PR, I think it makes sense to fix this instance now, and it'll only require changes to two more lines in this file.

cubedash/generate.py
Ariana-B: > Importing `cubedash.summary` imports 4 modules and all their dependencies (which is effectively the rest of Explorer), whereas importing `cubedash.summary._errors` imports a single module without any dependencies.

Yes, but `cubedash.summary._model` includes very few dependencies that `cubedash.summary._stores` doesn't already include, and `cubedash.summary._stores` imports `cubedash.summary._extents`, so at the end of the day I believe you wind up importing all the modules contained in `cubedash.summary.__init__` anyway.

cubedash/_filters.py
pjonsson: Fixed.

pjonsson:
Putting this in draft until I have some more time to work on it, I've split out the commit that moves `create_app` to `startup_utils` into #1108 for now.

PR #1113, opened 2026-04-19
index: use more specific type

pjonsson:
Change the type annotation
to a type that has
the alchemy_expression property.

This requires an ignore at
the callsite, but it's better
to get the correct type
information in most places.

Put this on top of a commit
that updates the dependencies
so we get the latest improvements
of type signatures from datacube-core.

----
📚 Documentation preview 📚: https://datacube-explorer--1113.org.readthedocs.build/en/1113/


cubedash/summary/_extents.py
```diff
@@ -497,8 +500,15 @@ def alchemy_expression(self):
         )
 
         # todo: look at grid_spec crs. Use it for defaults, conversion.
-        size_x, size_y = grid_spec.tile_size or (1000.0, 1000.0)
-        origin_x, origin_y = grid_spec.origin
+        # FIXME: grid_spec.tile_size and origin can be XY which is not iterable.
```
pjonsson, line 503: This is an old problem, so filed #1115 for this and put type ignores here for now.

cubedash/_utils.py
```diff
@@ -470,7 +470,8 @@ def datetime_from_metadata(dataset: Dataset) -> datetime:
         t = properties.get("datetime") or properties.get("dtr:start_datetime")
         return default_utc(dc_utils.parse_time(t))
     # stick with center time for EO datasets
-    return default_utc(dataset.center_time)
+    # FIXME: dataset.center_time can be None.
+    return default_utc(dataset.center_time)  # type:ignore[arg-type]
```
pjonsson, line 474: This is also an old problem so put in a type ignore and filed #1114 for this.

SpacemanPaul (approved):

PR #1122, opened 2026-04-23
tests: remove skipped test

pjonsson:
This test started being skipped
in 2021, so remove it now.

----
📚 Documentation preview 📚: https://datacube-explorer--1122.org.readthedocs.build/en/1122/


integration_tests/test_filter_geom.py
```diff
@@ -3,64 +3,13 @@
...
-        ValueError, match="No Shapely geometry can be created from null value"
-    ):
-        _ = shapely.ops.unary_union(
-            [ele.footprint_geometry.buffer(0.00) for ele in testing_polygon]
-        )
-
-    polygonlist = _polygon_chain(testing_polygon)
-    assert isinstance(polygonlist, list)
-    assert len(polygonlist) == 262
-    filtered_geom = _filter_geom(polygonlist)
```
pjonsson, line 62: Orthogonal to this PR is the question whether the code this tests is still relevant. It's the third fallback for trying to make sense of broken geometries as far as I understand, but it would be much easier to make datacube validate the input geometry and refuse to add datasets with broken geometries. That could then simplify the footprint calculation in Explorer to just skip datasets with broken geometries.

Ariana-B: Explorer geometries are kind of their own beast to a degree. For instance, telemetry datasets are non-geospatial as far as datacube-core is concerned, but Explorer includes logic to calculate their geometries. Furthermore, Explorer's use of a common grouping epsg for all datasets will break geometries that cannot be correctly reprojected to it (there is a desire to change the grouping epsg behaviour, but it would require a major refactor). Perhaps there is some old logic that can be cleaned up, but I don't think the solution you propose is likely to be sufficient.

pjonsson: Ok, that complicates the issue. Is there a telemetry dataset somewhere in the tests for Explorer? If not, could you add one so we get some increased coverage of the fallback paths?

And this PR just removes a test we haven't been running for the past 5 years, so no logic with how Explorer currently behaves or is tested is changed from merging this.

Ariana-B: > Is there a telemetry dataset somewhere in the tests for Explorer?

There are ls8 and ls7 satellite telemetry datasets in the test suite, but they're buried somewhere in `pq_count_summary.yaml.gz`.

Ariana-B (approved):
I wonder if it would make sense to move the lone remaining test into `test_model.py` and get rid of this test file altogether.

PR #1125, opened 2026-04-27
utils: fix type error

### LINKED ISSUE #1114: datetime_from_metadata: dataset.center_time can be None
The `datetime_from_metadata` function always returns a `datetime` according to the type annotation, but `dataset.center_time` can be `None`, in which case `default_utc` will crash.

pjonsson:
Make the second caller handle the None
case the same way as the first caller.

Fixes #1114

----
📚 Documentation preview 📚: https://datacube-explorer--1125.org.readthedocs.build/en/1125/

### DIFF
--- a/cubedash/_filters.py
+++ b/cubedash/_filters.py
@@ -43,18 +43,22 @@ bp = Blueprint("filters", __name__)
 
 
 @bp.app_template_filter("printable_time")
-def _format_datetime(date):
-    return date.strftime("%Y-%m-%d %H:%M:%S")
+def _format_datetime(date: datetime | None) -> str:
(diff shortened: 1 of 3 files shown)


Ariana-B (changes requested):
`datetime_from_metadata` is also called by the "metadata_center_time" filter, and its result passed into other filters ("localised_metadata_center_time" and "printable_time") which are not equipped to handle a null value.

pjonsson:
> `datetime_from_metadata` is also called by the "metadata_center_time" filter, and its result passed into other filters ("localised_metadata_center_time" and "printable_time") which are not equipped to handle a null value.

My PyCharm must have gotten confused somehow because it clearly showed a single caller besides the test, thanks for catching that!

Ariana-B (approved):

PR #1129, opened 2026-04-29
Improve stac fields extension logic

Ariana-B:
Rework the `_handle_fields_extension` logic to better follow the include/exclude semantics rules as described in the spec, particularly when it comes to handling nested fields.

----
📚 Documentation preview 📚: https://datacube-explorer--1129.org.readthedocs.build/en/1129/


integration_tests/test_stac.py
```diff
@@ -1412,6 +1412,39 @@ def test_stac_fields_extension(stac_client: FlaskClient) -> None:
...
     assert {"datetime"} == set(properties.keys())
 
+    # only include minimum addiitional fields needed to return a valid item
+    rv = stac_client.get(
+        "/stac/search?collection=ga_ls8c_ard_3&limit=5&fields=assets.nbart_red"
+    )
+    assert rv.status_code == 200
+    doc = rv.json
+    assert doc is not None, "Empty response from server"
+    assert doc.get("features")
```
pjonsson, line 1467: I realize the tests have quite a few complaints from MyPy, but doesn't this introduce more complaints? I would expect assigning to a variable and asserting on the None-ness of that to avoid that:
```python
features = doc.get("features")
assert features is not None
```

Edit: apparently MyPy is happy with the current formulation, but we're not type checking the tests in the CI, so I made a PR that enables type checking.

cubedash/_stac.py
```diff
@@ -659,64 +675,62 @@ def _handle_fields_extension(items: Sequence[Item], fields: dict) -> Sequence[It
...
+        "bbox",
+        "links",
+        "assets",
+        "stac_version",
+        # while not necessary for a valid stac item, we still want them included
+        "stac_extensions",
+        "collection",
+    }
+    include_set = (
+        set(fields.get("include")) if fields.get("include") else default_fields  # type: ignore[arg-type]
```
pjonsson: I'm guessing MyPy can't deduct that set() won't get `None`, if you rewrite to `set(includes) if (includes := fields.get("include")) else default_fields` it should be able to figure that out. And that avoids performing an unnecessary lookup in fields as well.

cubedash/_stac.py
```diff
@@ -659,64 +675,62 @@ def _handle_fields_extension(items: Sequence[Item], fields: dict) -> Sequence[It
...
+    )
+    exclude_set = set(fields.get("exclude")) if fields.get("exclude") else set()  # type: ignore[arg-type]
```
pjonsson: Same comment as previous line.

cubedash/_stac.py
```diff
@@ -645,12 +645,28 @@ def next_page_url(next_offset):
...
-def _get_property(prop: str, item: Item, no_default: bool = False):
+def _get_field(field: str, item: dict, no_default: bool = False) -> Any | None:
     """So that we don't have to keep using this bulky expression"""
-    return dicttoolz.get_in(prop.split("."), item.to_dict(), no_default=no_default)
+    return dicttoolz.get_in(field.split("."), item, no_default=no_default)
 
 
-def _handle_fields_extension(items: Sequence[Item], fields: dict) -> Sequence[ItemLike]:
+def _has_field(field: str, item: dict) -> bool:
+    try:
```
pjonsson, line 654: This function is called from filter() which is called inside a loop, and exceptions aren't terribly performant in any language. Does `item` contain any fields with the value `None`, or can the function be implemented as `return _get_field(field, item, False) is not None`?

cubedash/_stac.py
```diff
@@ -645,12 +645,28 @@ def next_page_url(next_offset):
...
+        _get_field(field, item, True)
+        return True
+    except KeyError:
+        return False
+
+
+def _set_field(field: str, filtered: dict, item: dict) -> dict:
+    if len(field.split(".")) > 1:
```
pjonsson, line 662: I don't know dicttoolz, but this function makes it look like `assoc_in` doesn't handle the base case where `field.split(".")` doesn't contain a period and returns `[field]`. Are the two branches really required?

cubedash/_stac.py
```diff
@@ -659,64 +675,62 @@ def _handle_fields_extension(items: Sequence[Item], fields: dict) -> Sequence[It
...
+        # the minimum number of additional fields needed for a valid stac item
+        # otherwise, start with all available fields for the item
+        if "include" in fields:
+            filtered_item: dict[str, Any] = {}
+            # filter out fields that don't have values in the item
+            include = set(filter(lambda i: _has_field(i, item_dict), include))
+            for i in include:
+                filtered_item = _set_field(i, filtered_item, item_dict)
+            # ensure all default fields are included without overwriting any nested fields in 'include'
+            missing_default = default_fields - set(filtered_item.keys())
```
pjonsson: Unlike `include`, this variable appears to only be used in the next line, but I had to do quite a bit of careful reading to make sure I hadn't missed anything.

I'm not sure what the exact cause is, but reading this code I'm having trouble following what value ends up in which variable at the end. If you inline the single-use variables, I think it will be easier to follow the variables that are re-assigned.

cubedash/_stac.py
Ariana-B: They're not, but my understanding is that `assoc` is faster than `assoc_in` so I thought it might be better to only use the latter when required - but it's possible any performance gain this affords is negligible.

cubedash/_stac.py
Ariana-B: As per the [stac spec](https://github.com/radiantearth/stac-spec/blob/master/overview.md#foundations):
> Note: Setting a field in JSON to `null` is not equivalent to a field not appearing in STAC, as JSON Schema tools treat them differently. STAC defines `null` explicitly for some fields, where it has a particular meaning. So `null` should not be used unless the STAC spec defines its use - instead the field should be left out entirely.

I'm not sure exactly which fields are explicitly nullable, so I wanted to err on the side of caution.

cubedash/_stac.py
Ariana-B: I did feel there was a more pythonic way to write this but it just wasn't coming to me, thanks

cubedash/_stac.py
Ariana-B: I thought assigning a variable with a (somewhat) descriptive name would help readability, but if that's not the case I'm happy to move it inline.
I've done my best to make this logic as straightforward and well-documented as I can but I know it's a little tricky, so definitely let me know if there are comments I need to improve/add.

integration_tests/test_stac.py
Ariana-B: The current assertion also verifies that `features` is not empty, which is equally important since I index into it.

cubedash/_stac.py
```diff
@@ -659,64 +675,59 @@ def _handle_fields_extension(items: Sequence[Item], fields: dict) -> Sequence[It
...
+        "geometry",
+        "bbox",
+        "links",
+        "assets",
+        "stac_version",
+        # while not necessary for a valid stac item, we still want them included
+        "stac_extensions",
+        "collection",
+    }
+    include_set = set(inc) if (inc := fields.get("include")) else default_fields
```
pjonsson, line 692: Is the logic implemented by this function the same as this [table](https://github.com/stac-api-extensions/fields/blob/main/README.md#null-vs-empty-vs-missing) except that it always returns a valid STAC item?

cubedash/_stac.py
```diff
@@ -659,64 +675,59 @@ def _handle_fields_extension(items: Sequence[Item], fields: dict) -> Sequence[It
...
-
-        # datetime is one of the default fields, but might be included as start_datetime/end_datetime instead
-        if _get_property("properties.start_datetime", item) is None:
-            dt_field = ["properties.start_datetime", "properties.end_datetime"]
+        item_dict = item.to_dict()
+        include, exclude = include_set, exclude_set
+        # datetime is a default field, but might be included as start_datetime/end_datetime
+        # stick with just 'datetime' if both it and the start/end pair are present
+        if _get_field("properties.datetime", item_dict) is not None:
+            include.add("properties.datetime")
```
pjonsson, line 701: I think this modifies `include_set`, so the next iteration with a different item might take the other branch and suddenly include contains all three values.

integration_tests/test_stac.py
```diff
@@ -1412,6 +1412,39 @@ def test_stac_fields_extension(stac_client: FlaskClient) -> None:
     properties = doc["features"][0]["properties"]
     assert {"datetime"} == set(properties.keys())
 
+    # only include minimum addiitional fields needed to return a valid item
+    rv = stac_client.get(
```
pjonsson, line 1461: I don't know what the current tests do, but it's probably wise to ensure that each bullet from the [list](https://github.com/stac-api-extensions/fields/blob/main/README.md#includeexclude-semantics) is covered by some test. Especially
> If the same field is present in both `include` and `exclude`, it should be included.

seems like something that is easy to break when updating the code.

cubedash/_stac.py
```diff
@@ -659,64 +675,59 @@ def _handle_fields_extension(items: Sequence[Item], fields: dict) -> Sequence[It
...
-                    func=lambda _: _get_property(inc, item, no_default=True),
-                )
-            except KeyError:
-                continue
+            include.update(["properties.start_datetime", "properties.end_datetime"])
+
+        # if 'include' is present at all, include only the specified fields, plus
+        # the minimum number of additional fields needed for a valid stac item
+        # otherwise, start with all available fields for the item
+        if "include" in fields:
```
pjonsson, line 708: I think there are some cases missing from this condition (at least to cover the first and fourth line of the [table](https://github.com/stac-api-extensions/fields/blob/main/README.md#null-vs-empty-vs-missing)).

cubedash/_stac.py
Ariana-B: Yes, that's the goal

integration_tests/test_stac.py
Ariana-B: Most of those are already covered, but I've added tests for those that weren't, and more explicit coverage for a couple others.

cubedash/_stac.py
Ariana-B: The first line of the table is already covered by the tests and I've added a test for the fourth line with no issue, so unless there's something else I missed this condition is covering all cases.

cubedash/_stac.py
Ariana-B: Ugh, I keep forgetting that's how python behaves. Thanks for pointing it out

integration_tests/test_stac.py
pjonsson: Could you update the comments for the tests with a stable link to the table and a reference to the row for each test, perhaps something like this: 
```python
# The following tests ensure that Explorer complies with each row of 
# https://github.com/stac-api-extensions/fields/blob/d0e1994f0508d45ca6da9df7a7aa41e02f4cd002/README.md?plain=1#L122
#
# Row 1, exclude without include ...
...
assert ...
...
# Row 2, ..
...
assert ...
...
# Row 3, ...
```
It doesn't simplify the implementation of the specification, but at least it makes it clear that the tests are "exhaustive" so if we trust the tests we can trust the implementation.

Ariana-B: The code already contains a link to the stac spec, and each of the tests already has a comment describing the behaviour it is testing.

cubedash/_stac.py
pjonsson: Not sure what I was thinking when I wrote that comment.

I think it's the distinction between missing, empty and `null` that is confusing me, how is `include: null` encoded in the URL in the tests, and what does `fields.get("include")` return in the code here when we get null?

Edit: if there aren't tests that inputs null for the first and second row of the table, I think it would be nice to add those.

integration_tests/test_stac.py
pjonsson: Yes, but prepending "Row X, " to each comment I can just check that the first test's comment starts with "Row 1, " and the last test's comment starts with "Row 8, " to know that the table is covered exhaustively.

cubedash/_stac.py
Ariana-B: `fields.get("include")` is falsy whether the value is `null` or empty, so in both cases `include` will be set to the default fields - this is covered by test cases 3, 5, and 9.
The missing case is handled by the else case here (i.e. `"include" not in fields`), and is covered in my second test case.

pjonsson:
> Rework the `_handle_fields_extension` logic to better follow the include/exclude semantics rules as described in the spec, particularly when it comes to handling nested fields.

"Better" gives quite a lot of leeway with how spec compliant Explorer is with this PR, but you mentioned that the table is exhaustively covered by the tests, does that mean this PR makes Explorer fully spec compliant in this area?

If you add the "Row 1, " comments to the tests so it's obvious for a reader that the table is exhaustively covered by the tests I think this PR can be merged.

Ariana-B:
> "Better" gives quite a lot of leeway with how spec compliant Explorer is with this PR, but you mentioned that the table is exhaustively covered by the tests, does that mean this PR makes Explorer fully spec compliant in this area?

I'm leaving that leeway because Explorer *should* now be technically fully compliant with the spec, however there are recommended behaviours that it does not currently implement (specifically relating to returning invalid STAC Items - this is documented in the `_handle_fields_extension` docstring), so I don't want to discount the possibility of future improvements to further align the implementation with the spec recommendations. Similarly, the include/exclude semantics rules should now be fully covered by the tests, but it's always possible there are edge cases beyond what is explicitly documented in the spec that I haven't considered.

pjonsson:
> I'm leaving that leeway because Explorer _should_ now be technically fully compliant with the spec, however there are recommended behaviours that it does not currently implement (specifically relating to returning invalid STAC Items - this is documented in the `_handle_fields_extension` docstring), so I don't want to discount the possibility of future improvements to further align the implementation with the spec recommendations. 

From a development perspective, if you get a new job and someone else looks at this PR in 3 years to try to figure out how Explorer arrived at the current state, that person will have to read everything in this PR to figure out what the state was meant to be after merging this PR (=effectively answering the third bullet of https://who-t.blogspot.com/2009/12/on-commit-messages.html). And the only reason they don't have to infer that is because I happened to ask.

Your answer is a much stronger statement than the PR description and it's your accomplishment, be proud of what you've done!

> Similarly, the include/exclude semantics rules should now be fully covered by the tests, but it's always possible there are edge cases beyond what is explicitly documented in the spec that I haven't considered.

There is always a risk that one missed something in the implementation (or that something goes wrong in the execution) causing Explorer to not behave as intended. If this implementation has a bug, I also missed it during the code review.

And on reviewing, the stronger statement aids reviewing since it's more specific, knowing that Explorer is supposed to be compliant except for X and Y makes it possible to refute the statement if I happen to find Z. Maybe that was just missed/forgotten when writing the description so it's an easy fix to update the description, or it's a real bug discovered and that needed to be fixed. In the end we all come out as winners, either you get more knowledgeable collaborators which will help in the long run, or you get the chance to fix the bug immediately when you still have all the details fresh in your mind and don't have to revisit it in 6 months.

I don't think anyone reasonable expects completely bug free software, but it's really nice if when we're 3 years down the line don't have to guess whether the behavior is intentional or not.

pjonsson (approved):

Ariana-B:
> And on reviewing, the stronger statement aids reviewing since it's more specific, knowing that Explorer is supposed to be compliant except for X and Y makes it possible to refute the statement if I happen to find Z. Maybe that was just missed/forgotten when writing the description so it's an easy fix to update the description, or it's a real bug discovered and that needed to be fixed. 

I didn’t rehash the specifics of Explorer’s implementation in the PR description since the docstring already describes the ways in which ways Explorer is not in line with the recommended behaviour outlined in the spec. But if that documentation is unclear, by all means let me know what can be clarified and I will update it.

For the record, my understanding is that my original handling of nested fields, while not conforming to what’s described in the include/exclude table, was actually already compliant with the spec:
> Implementations are also not required to implement semantics for nested values whereby one can include an object, but exclude fields of that object, e.g., include `properties` but exclude `properties.datetime`.

PR #1130, opened 2026-04-29
Type check integration_tests

pjonsson:
Add a type alias for the two different
indices that Explorer supports, and
use that in the tests. Besides two
complaints from MyPy, the rest of
the tests type check, so add two
type ignores and enable type checking
in the lint workflow.

----
📚 Documentation preview 📚: https://datacube-explorer--1130.org.readthedocs.build/en/1130/

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -35,7 +35,7 @@ jobs:
         python-version: ["3.10"]
         lintcommand:
           - "ruff check --output-format=github"
-          - "mypy --follow-imports=silent cubedash/*.py cubedash/index/*.py cubedash/startup_utils cubedash/summary cubedash/testutils"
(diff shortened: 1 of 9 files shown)


cubedash/summary/_extents.py
```diff
@@ -50,6 +52,7 @@
     Path(__file__).parent.parent / "data" / "WRS2_ascending" / "WRS2_acsending.shp",
 ]
 
+DatacubeIndex: TypeAlias = PgresIndex | PgisIndex
```
Ariana-B: This strikes me as a very odd place to define this type alias since it is never used in this file.
On the other hand, `summary/_stores`, `summary/show`, and `generate` already contain the double index imports - `_stores` would probably be the most logical place to define the type alias.

pjonsson: I put it where we had some similar type aliases already, but I've moved it to stores as you suggest.

Ariana-B (approved):

PR #1131, opened 2026-04-29
pyproject: add missing pandas dependency

pjonsson:
This package is used by summarise.

----
📚 Documentation preview 📚: https://datacube-explorer--1131.org.readthedocs.build/en/1131/


Ariana-B (approved):
I think we may also need to add werkzeug to our dependencies.

pjonsson:
> I think we may also need to add werkzeug to our dependencies.

You're right, added that in a second commit!

PR #1159, opened 2026-06-02
Catch query timeouts

SpacemanPaul:
Support Query timeouts, as available from ODC 1.9.20.

1) Refactor and cleanup of Explorer index layer to use new execution methods exposed in core driver layer.
2) Add error handler for QueryTimeout error
3) Bypass the `inject_globals` context handler when rendering 5xx error pages.

Tests pass for the refactor and cleanup.  The current database test utils are not well adapted to testing timeouts - I need to think a bit more about how to test systematically.

Also forgot to run mypy - I knew that was going to be "fun".

----
📚 Documentation preview 📚: https://datacube-explorer--1159.org.readthedocs.build/en/1159/


pjonsson:
This is a big PR, can you cherry-pick the first two commits that just refactors things without changing any functionality and put them in a separate PR so it's easier to review those in isolation first?

SpacemanPaul:
> This is a big PR, can you cherry-pick the first two commits that just refactors things without changing any functionality and put them in a separate PR so it's easier to review those in isolation first.

Good call.

PR #1161, opened 2026-06-02
Prevent inject_globals being called on 500 errors.

SpacemanPaul:
500 errors do not render the full template and do not need globals injected.

Add a util method to render directly with Jinja, bypassing Flask's context processors.

----
📚 Documentation preview 📚: https://datacube-explorer--1161.org.readthedocs.build/en/1161/

### DIFF
--- a/cubedash/_utils.py
+++ b/cubedash/_utils.py
@@ -43,6 +43,7 @@ if TYPE_CHECKING:
 
     from datacube.index.fields import Field
     from datacube.model import Dataset, Product
+    from flask import Flask
     from odc.geo import Geometry
     from shapely.geometry.base import BaseGeometry
     from werkzeug.datastructures import MultiDict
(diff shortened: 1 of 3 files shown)


cubedash/_utils.py
```diff
@@ -96,6 +96,13 @@ def render(template, **context):
     return render_template(template, **context)
 
 
+def render_without_context_processors(app, template, **context):
+    # When e.g. rendering error pages, we don't want to trigger the context processors
+    # Use native Jinja template rendering instead (bypassing flask)
+    template = app.jinja_env.get_template(template)
```
pjonsson: Just `return app.jinja_env.get_template(template).render(**context)` to avoid the re-assignment of template (and giving the template variable multiple types).

cubedash/_utils.py
```diff
@@ -96,6 +96,13 @@ def render(template, **context):
     return render_template(template, **context)
 
 
+def render_without_context_processors(app, template, **context):
```
pjonsson: I'm not sure if we have anything useful for types of the generic stuff, but at least `app` should be of type `flask.Flask`, right?

cubedash/_utils.py
```diff
@@ -43,6 +43,7 @@
 
     from datacube.index.fields import Field
     from datacube.model import Dataset, Product
+    from flash import Flask
```
pjonsson: Typo?

pjonsson (approved):

PR #1162, opened 2026-06-02
Catch timeouts

SpacemanPaul:
Handle Timeouts

includes the refactor in #1163 - in draft until that is merged, otherwise ready for review.

- ensure that all query timeouts will be caught
- handle QueryTimeout exceptions at the flask level.

Configure Flask to handle QueryTimeout exceptions.

----
📚 Documentation preview 📚: https://datacube-explorer--1162.org.readthedocs.build/en/1162/

### DIFF
--- a/cubedash/index/postgis/_api.py
+++ b/cubedash/index/postgis/_api.py
@@ -5,7 +5,12 @@ from typing import Any
 import shapely.ops
 import structlog
 from cachetools.func import lru_cache
-from datacube.drivers.common_psql import as_role, create_schema, has_roles
+from datacube.drivers.common_psql import (
+    as_role,
+    catch_timeout,
+    create_schema,
+    has_roles,
+)
 from datacube.drivers.postgis._api import PostgisDbAPI
 from datacube.drivers.postgis._fields import SimpleDocField
 from typing_extensions import override
@@ -814,7 +819,7 @@ class ExplorerIndex(ExplorerAbstractIndex):
     def spatial_select_query(
         self, clauses: Sequence[Label | ClauseElement], full: bool = False
     ) -> Select:
-        query = select(*clauses)
+        query = select(*clauses)  # type: ignore[arg-type]
         if full:
             return query.select_from(DatasetSpatial).join(
                 ODC_DATASET, onclause=ODC_DATASET.id == DatasetSpatial.id
@@ -838,6 +843,7 @@ class ExplorerIndex(ExplorerAbstractIndex):
             )
 
     @override
+    @catch_timeout
     def schema_compatible_info(self, for_writing_operations_too=False):
         """
         Schema compatibility information
@@ -855,6 +861,7 @@ class ExplorerIndex(ExplorerAbstractIndex):
             )
 
     @override
+    @catch_timeout
     def create_schema(self) -> bool:
         # Ensure ODC roles exist (roles were formerly optional but not using them is now deprecated)
         with self.engine.connect() as conn:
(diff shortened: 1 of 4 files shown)


pjonsson (dismissed):
There is a reasonably large change in this PR that updates how connections are used, and the types for the abstract class and its subclasses.

I don't doubt that you got it right since you made the changes, but could you split that part out into its own PR to make it easier to review?

cubedash/summary/_stores.py
```diff
@@ -459,7 +459,10 @@ def refresh_product_extent(
         if product.id is None:
             return None
         # if change_count or force_dataset_extent_recompute:
-        earliest, latest, total_count = self.e_index.product_time_overview(product.id)
+        _oview = self.e_index.product_time_overview(product.id)
+        if _oview is None:
+            raise ValueError(f"No product summary for {product.name!r}")
```
pjonsson: I think the `None` return type of this function is for failure, so this should return that.

cubedash/summary/_summarise.py
```diff
@@ -70,11 +70,11 @@ def calculate_summary(
...
 
-        rows = self.e_index.srid_summary(where_clause).fetchall()
-        log.debug("summary.query.done", srid_rows=len(rows))
-
-        assert len(rows) == 1
-        row = dict(rows[0]._mapping)
+        _row = self.e_index.srid_summary(where_clause)
+        log.debug("summary.query.done")
+        if _row is None:
+            raise ValueError(f"No summary found for {product_name} {time}")
```
pjonsson: Add a `| None` to the return type instead and handle the possibility of failure in the caller(s), getting an error here and raising an exception will just make things crash in a different place than they used to.

cubedash/index/postgis/_api.py
```diff
@@ -443,23 +455,24 @@ def collections_search_query(
...
             if latest_arrival_date is None:
                 raise EmptyDbError()
 
             datasets_since_date = (latest_arrival_date - period_length).date()
 
             # shouldn't this be getting from odc.dataset combined with dataset_spatial?
             # no point returning datasets that have been added in the odc database but not the cubedash one
-            return conn.execute(
-                text("""
+            return list(
```
pjonsson: Doesn't `fetchall()` already have a return type of `Sequence[something]`?

cubedash/index/postgis/_api.py
```diff
@@ -739,8 +752,8 @@ def upsert_datasets(
         column_values: dict[str, Label],
         after_date: datetime | None,
     ) -> int:
-        column_values["id"] = ODC_DATASET.id
-        column_values["product_ref"] = ODC_DATASET.product_ref
+        column_values["id"] = ODC_DATASET.id  # type: ignore[assignment]
```
pjonsson: Nothing was changed in this method, why are the type ignores suddenly required?

SpacemanPaul:
> There is a reasonably large change in this PR that updates how connections are used, and the types for the abstract class and its subclasses.
> 
> I don't doubt that you got it right since you made the changes, but could you split that part out into its own PR to make it easier to review?

I'm not sure what will be left, but I'll see if I can break it down further.

Edit, refactor moved to #1163  this PR can be merged after that.

cubedash/index/postgis/_api.py
SpacemanPaul: Good question.  There were a few mypy errors that made no sense to me.

Ariana-B (approved):

PR #1163, opened 2026-06-02
Connection usage cleanup and API refactor.

SpacemanPaul:
Another split off from the original monolithic catch-timeout PR.

This one just includes the refactor of the abstract API interface and cleanup of connection usage.

----
📚 Documentation preview 📚: https://datacube-explorer--1163.org.readthedocs.build/en/1163/


cubedash/index/postgis/_api.py
```diff
@@ -787,7 +791,7 @@ def synthesize_dataset_footprint(
                     }
                     for id_, sat_path, sat_row in rows
                 ],
-            )
+            ).rowcount
```
Ariana-B, line 794: The result of this is never used so there's not much point in changing its return value/type.

Ariana-B (approved):

cubedash/index/postgis/_api.py
SpacemanPaul: Equally not much point in leaving it how it is.  May as well return something that might potentially be useful in future.

pjonsson: The rowcount is often not useful in psycopg3.

I don't know the details, but I'm guessing it's a change done to improve performance in some respect.

PR #1164, opened 2026-06-03
Remove recursive query from raw metadata.

SpacemanPaul:
- don't call get with include_sources
- use new core methods to do simple shallow read of lineage for raw dataset metadata
- Always write EO3-style id-only lineage on EO-format datasets in raw dataset metadata

----
📚 Documentation preview 📚: https://datacube-explorer--1164.org.readthedocs.build/en/1164/

### DIFF
--- a/cubedash/_utils.py
+++ b/cubedash/_utils.py
@@ -815,13 +815,6 @@ EODATASETS_PROPERTY_ORDER = [
     "lineage",
     "product_flags",
 ]
-EODATASETS_LINEAGE_PROPERTY_ORDER = [
-    "algorithm",
-    "machine",
-    "ancillary_quality",
-    "ancillary",
-    "source_datasets",
-]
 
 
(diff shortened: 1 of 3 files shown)


cubedash/_utils.py
```diff
@@ -721,25 +728,6 @@ def get_property_priority(ordered_properties: list, keyval):
         )
     )
 
-    # Order any embedded ones too.
-    if "lineage" in ordered_metadata:
-        ordered_metadata["lineage"] = dict(
-            sorted(
-                ordered_metadata["lineage"].items(),
-                key=functools.partial(
-                    get_property_priority, EODATASETS_LINEAGE_PROPERTY_ORDER
```
Ariana-B, line 730: Should be able to delete `EODATASETS_LINEAGE_PROPERTY_ORDER`

Ariana-B (approved):

PR #1173, opened 2026-06-07
pyproject: require Python 3.12

pjonsson:
Python 3.10 is reaching end of life
at the end of October this year
and it's getting increasingly harder
to keep the support going, so bump
the required Python to 3.12.

----
📚 Documentation preview 📚: https://datacube-explorer--1173.org.readthedocs.build/en/1173/

### DIFF
--- a/cubedash/_stac.py
+++ b/cubedash/_stac.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import json
 import uuid
-from datetime import datetime, timedelta, timezone
+from datetime import UTC, datetime, timedelta
 from datetime import time as dt_time
 from functools import partial
 from typing import Any, TypeAlias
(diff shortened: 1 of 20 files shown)


cubedash/_stac.py
```diff
@@ -86,8 +86,8 @@ def dissoc_in(d: dict, key: str):
 
 def utc(d: datetime):
```
pjonsson, line 87: We have multiple copies of these functions, but that is not an issue from this PR, so let's sort that out after this is merged.

pjonsson:
Rebased against latest develop, and added a commit on top that makes the CI run MyPy on more files.

Also discovered that we should be depending on the backend-sqlalchemy optional of pygeofilter which has nothing to do with Python 3.12, but since that changes uv.lock I added the commit here to avoid merge conflicts just before the release.

cubedash/index/postgis/_api.py
```diff
@@ -760,8 +759,8 @@ def upsert_datasets(
         column_values: dict[str, Label],
         after_date: datetime | None,
     ) -> int:
-        column_values["id"] = ODC_DATASET.id
-        column_values["product_ref"] = ODC_DATASET.product_ref
+        column_values["id"] = ODC_DATASET.id  # type: ignore[assignment]
+        column_values["product_ref"] = ODC_DATASET.product_ref  # type: ignore[assignment]
         only_where = [
```
SpacemanPaul, line 764: Ha - these were the type hints I added and then removed on the big PR that got split into small ones.

SpacemanPaul (approved):

PR #1182, opened 2026-06-19
Documentation Theme/Appearance/Tidying

omad:
Mostly theme/appearance/tidying, laying the groundwork for fleshing out some guidance on actually running Explorer in production.

Screenshot 2026-06-19 at 5 52 17 pm

----
📚 Documentation preview 📚: https://datacube-explorer--1182.org.readthedocs.build/en/1182/


CONTRIBUTING.rst
```diff
@@ -45,7 +43,7 @@ and "help wanted" is open to whoever wants to implement it.
 Write Documentation
 ~~~~~~~~~~~~~~~~~~~
 
-datacube-explorer could always use more documentation, whether as part of the
+ODC Explorer could always use more documentation, whether as part of the
```
pjonsson: I don't know if we should rename the project, but the next line immediately uses "datacube-explorer" as name.

docs/conf.py
```diff
@@ -70,56 +72,54 @@
...
+    # "navbar_align": "left",
+    "show_prev_next": False,
+    "collapse_navigation": True,
+    "use_edit_page_button": True,
+    # "footer_items": ["odc-footer"],
+    "secondary_sidebar_items": ["page-toc", "edit-this-page"],
+    "icon_links": [
+        {
+            "name": "GitHub",
+            "url": "https://github.com/opendatacube/datacube-core",
```
pjonsson: Do you want to link core here, or was the intention to link explorer?

CONTRIBUTING.rst
omad: There's a whole lot more wrong than that in this contributions document. I didn't look too hard, and felt I was churning too much already.

Happy to change all the occurances of "ODC Explorer" back to "Datacube Explorer" though, if you want, I didn't really set out to rename the project :)

docs/conf.py
omad: Whoops, copy/paste mistake! Thanks

docs/config.rst
```diff
@@ -1,204 +1,232 @@
...
 
-.. py:data:: SHOW_DATA_LOCATION
+.. confval:: SHOW_DATA_LOCATION
 
     S3 buckets for which to return a browseable bucket link instead of the plain S3 link
 
-    Default: ``{}``
-    Example: ``{ 'dea-public-data': 'data.dea.ga.gov.au'}``
+    :default: ``{}``
+    :type: ``dict[str, str]```
```
pjonsson: I'm guessing it doesn't need to be mutable, should this be `Mapping[str, str]`?

omad: Sure... it's just supposed to be a hint though, it's not like this is in any way linked to a type checker.

CONTRIBUTING.rst
omad: Okay, the name change has been undone. I found the templated `datacube-explorer` offensive and got carried away.

CONTRIBUTING.rst
```diff
@@ -1,17 +1,15 @@
-How to contribute to Datacube-explorer
-======================================
+How to contribute to Dataube Explorer
```
pjonsson: Missing a "c" in "Datacube".

docs/conf.py
```diff
@@ -162,83 +162,19 @@
 # to template names.
 # html_additional_pages = {}
 
-# If false, no module index is generated.
```
pjonsson, line 165: It might be useful to run the command that re-generates `conf.py` when upgrading Sphinx and then rolling back the parts where we have changed the configuration, are the removed lines things that the current version of Sphinx doesn't generate?

docs/debugging.rst
```diff
@@ -9,9 +9,9 @@ a production environment.** The debugger allows executing arbitrary
 Python code from the browser. It's protected by a pin, but that should
 not be relied on for security.
 
-Use an error logging tool, such as Sentry, as described in
-:ref:`error-logging-tools`, or enable logging and notifications as
-described in :doc:`/logging`.
+Use an error logging tool, such as Sentry, as described below
+, or enable logging and access them via your system or Kubernetes log tool
```
pjonsson: Put the comma at the end of the previous line.

docs/deploying.rst
```diff
@@ -97,6 +100,9 @@ Sentry error reporting is supported and can be setup as per :any:`sentry-env`
 Roles for production deployments
 ---------------------------------
 
+.. todo:: The way database roles has changed in 2026 and this information is no longer
```
pjonsson: Explorer uses the core roles, so link that part of the core documentation.

docs/index.rst
```diff
@@ -1,34 +1,36 @@
-Welcome to datacube-explorer's documentation!
-=============================================
+Dataube Explorer
+============
```
pjonsson: I think you get warnings from Sphinx (or something underneath Sphinx) when the heading level indicator is not the same amount of characters as the heading above it.

docs/index.rst
```diff
@@ -1,34 +1,36 @@
...
+   operating
 
-
-Additional Notes
-----------------
+Development Guide
+-----------------
 
 .. toctree::
+   :caption: Developing
```
pjonsson: There's two headings called "Developing" under "Development Guide" for some reason, perhaps it's this line causing the second level to appear?

pjonsson:
Can you do the same with enabling fail on warnings here that I did in core? (Reference commit:  https://github.com/opendatacube/datacube-core/commit/cb7efc36c0df3b2743858706180eb1671c8ecffb)

docs/index.rst
omad: Thanks. It's a bit weird how it works. I've tidied it up the best I can for the time being.

I want to use the same `pydata_sphinx_theme` as Core does. It generates the top navigation bar, and links within a "section", based upon having multiple `toctree`s in the index page.

The front docs page still has a title, and then the top navigation link in the toctree is pretty much a duplicate, but, I think it's fine for now and it would be a lot of messing around to avoid.

omad:
> Can you do the same with enabling fail on warnings here that I did in core? (Reference commit: [opendatacube/datacube-core@cb7efc3](https://github.com/opendatacube/datacube-core/commit/cb7efc36c0df3b2743858706180eb1671c8ecffb))

Done

docs/conf.py
omad: I'm not aware of a command for doing that. The best you can do is use `sphinx-quickstart` and manually fill in details for a completely new documentation site.

The configuration I removed was for different build targets link PDF via tex that we have never used anyway and just make for an unwieldy configuration.

pjonsson: Yeah, `sphinx-quickstart` was probably what I used when I shifted the documentation source directory of core. I don't recall the details, but I assume I removed the existing configuration, ran sphinx-quickstart to get a new config generated for the current version of Sphinx, and then I used the diff-viewer in PyCharm to revert the changes to the parts of the configuration that I wanted to remain on a line-by-line basis. (I'm sure there's something similar to PyCharm's diff-viewer in VSCode and any other IDE, and `git add -p` and splitting large changes can probably do more or less the same as well.)

So removing the parts of the config file that we don't use will give a larger diff if we need to update the config for a more recent version of Sphinx sometime in the future.

CONTRIBUTING.rst
pjonsson: It's missing a "c" in all the other places as well.

SpacemanPaul (dismissed):
Mostly looks good to me, except for the consistently missing "c" in "dataube", as previously noted by PJ.

CONTRIBUTING.rst
omad: Oh my, search and replace disaster! Sorry!

Fixed now.

docs/conf.py
omad: There shouldn't ever be a need to replace the Sphinx configuration. I've never had to in the 15 odd years I've been using it. Every now and then there's extra features added, but as often as not the default configuration doesn't include them anyway. The generated config is merely a starting point.

docs/index.rst
omad: reStructuredText is extremely picky about things like that, it's annoying, and one of the reasons Markdown beat it in the popularity stakes.

Should be fixed now.

pjonsson (approved):

PR #1228, opened 2026-08-24
cubedash-gen: catch database errors

pjonsson:
This avoids printing a long backtrace
to the user when the database isn't
reachable.

----
📚 Documentation preview 📚: https://datacube-explorer--1228.org.readthedocs.build/en/1228/


pjonsson:
This should pass the tests once #1229 is merged.

Edit: Nope, #1229 is not it. I have no idea why the tests suddenly stopped working, but it's not caused by this PR.

Ariana-B (approved):

PR #1235, opened 2026-09-01
stores: catch database errors

pjonsson:
Triggered this when updating
the index, so catch the error
and return an error instead
of crashing.

----
📚 Documentation preview 📚: https://datacube-explorer--1235.org.readthedocs.build/en/1235/


omad (approved):
thankyou!

PR #1238, opened 2026-09-02
[docker/gha] arm64+amd64, `:latest` = release, simplify CI

omad:
This updates and replaces the docker image build workflow in GitHub Actions. It uses the Docker GitHub Builder [1] reusable workflow, which makes it much easier to build multi-platform images efficiently in GHA across multiple runner architectures and then merge them, as well as better use of cache.

I've also updated the tags and labels that get applied to the image.

It should be backwards compatible, and maintain the semantics/versioning done by  setuptools-scm.

The expected **incompatible** change is to the `:latest` tag, which will now be applied to **released** versions, instead of development versions. This is inline with OWS and is the most common practice broadly.

----
📚 Documentation preview 📚: https://datacube-explorer--1238.org.readthedocs.build/en/1238/

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,74 +1,42 @@
-# Files that are only used by docker/docker compose.
-docker-compose.yaml
-docker-compose.*.yaml
-docker-compose.yml
-docker-compose.*.yml
-.docker
-.dockerignore
+# .dockerignore
+#
+# With BuildKit (this repo uses `# syntax=docker/dockerfile:1`), the build
+# context is transferred incrementally and only the files referenced by the
+# Dockerfile are sent, so an exhaustive ignore list is no longer needed for
+# efficiency. This keeps only the exclusions that genuinely matter:
+#
+#   - VCS metadata: large, and the image build deliberately avoids needing git
+#     (it uses SETUPTOOLS_SCM_PRETEND_VERSION), so `.git` must never be copied.
+#   - Local virtualenvs and caches: potentially huge and host-specific.
+#   - Local secrets / environment-specific files: must never leak into an image.
+#   - OS/editor cruft.
 
-# Byte-compiled / optimized / DLL files
-**/__pycache__
-**/*.py[cod]
-
-#Git/VCS
+# VCS
 .git
 .jj
 
-# CI / distribution / packaging
-**/*.egg-info
-.github
-.gitignore
-.mergify.yml
-.pre-commit-config.yaml
-.readthedocs.yaml
-.stylelintrc
-.venv
-.yamllint
-build
-dist
-Makefile
-pylintrc
+# Docker's own files (not needed inside the image)
+.dockerignore
 
-# Test and coverage reports
-**/htmlcov
+# Local virtualenvs and tool caches
+.venv
+**/__pycache__
+**/*.py[cod]
+**/.cache
+**/.mypy_cache
+**/.pytest_cache
+**/.ruff_cache
 **/.tox
+**/.hypothesis
 **/.coverage
 **/.coverage.*
-**/.cache
-**/nosetests.xml
-**/coverage.xml
-**/*.cover
-**/.hypothesis
-.mypy_cache
-.pytest_cache
-.ruff_cache
-conftest.py
-integration_tests
-tests
 
-# Django stuff:
-**/*.log
-
-# Documentation
-*.md
-*.png
-*.rst
-docs
-LICENSE
-
-# PyBuilder
-**/target
-
-# iPython Notebook
-**/.ipynb_checkpoints
-
-# Mac OS X
-**/.DS_Store
-
-# Environment-specific templates
+# Local secrets / environment-specific files
+**/*.env.py
 **/*.env.html
 **/settings.env.py
 
-# IDE settings
+# OS / editor cruft
+**/.DS_Store
 .idea
 .vscode
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -7,8 +7,6 @@ on:
   push:
     branches:
       - develop
-    paths:
-      - "**"
 
   release:
     types: [published]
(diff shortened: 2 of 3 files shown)


SpacemanPaul (approved):

.dockerignore
```diff
@@ -1,74 +1,42 @@
...
-# Byte-compiled / optimized / DLL files
-**/__pycache__
-**/*.py[cod]
-
-#Git/VCS
+# .dockerignore
+#
+# With BuildKit (this repo uses `# syntax=docker/dockerfile:1`), the build
+# context is transferred incrementally and only the files referenced by the
+# Dockerfile are sent, so an exhaustive ignore list is no longer needed for
```
pjonsson, line 5: I'm sure the incremental part the comment mentions is documented somewhere in the Docker documentation, but I'm doubtful that it works that way in practice because I'm fairly sure I've observed getting a multi-gigabyte context when building despite not copying those files into the image I was building. I don't have any hard evidence for this though, just a strong doubt.

Something I do have hard evidence about is the Docker cache though, my experience is that Docker hits the cache when it shouldn't, and it misses the cache when it shouldn't. I don't know what is causing this, but being forced to clear the Docker cache isn't an unusual thing. On top of that there are the race conditions, so Docker either loses track of blobs kept on disk and fills up all available space, or tries to access blobs that aren't available in storage and causing the build to fail.

Given this discrepancy between theory and practice with Docker, I don't think we should remove stuff from `.dockerignore` unless we actually want that thing in the image. Ignoring only works for files/directories we aren't interested in, but some is better than nothing.

I'm more than happy to add new things in this file if we've missed to add something.

Dockerfile
```diff
@@ -1,9 +1,9 @@
 # syntax=docker/dockerfile:1
-FROM ghcr.io/osgeo/gdal:ubuntu-small-3.13.3@sha256:64250faf833c06d4b21afce4c27190039ba7ab58d70f0eebc87cf77d929c0b40 AS base
 
-LABEL org.opencontainers.image.source=https://github.com/opendatacube/datacube-explorer
-LABEL org.opencontainers.image.description="Datacube Explorer"
-LABEL org.opencontainers.image.licences="Apache-2.0"
+# See https://reproducible-builds.org/docs/source-date-epoch/
+ARG SOURCE_DATE_EPOCH=1704067200
```
pjonsson, line 4: https://github.com/moby/buildkit/blob/master/docs/build-repro.md#source_date_epoch states:

> The buildctl CLI (>= v0.13) and Docker Buildx (>= 0.10) automatically propagate the $SOURCE_DATE_EPOCH environment value from the client host to the SOURCE_DATE_EPOCH build arg.

Do we really need a default value for this?

.github/workflows/docker.yml
```diff
@@ -26,95 +24,93 @@ concurrency:
...
+    permissions:
+      contents: read      # fetch the repository content
+      id-token: write     # sign attestations and cache entries via OIDC
+      packages: write     # push to GHCR
+      attestations: write # write the attestation storage record
+    with:
+      output: image
+      push: true
+      platforms: linux/amd64,linux/arm64
+      cache: true
```
pjonsson, line 94: Cache(-from) should be disabled for releases.

Even if the signed cache stuff mentioned by the reusable workflow documentation prevents cache-poisoning attacks (does it?), the Docker cache might hit when it shouldn't. For development builds this isn't the end of the world, but release builds shouldn't be up to luck.

.github/workflows/docker.yml
```diff
@@ -26,95 +24,93 @@ concurrency:
...
+  prepare:
     runs-on: ubuntu-latest
-    if: github.ref == 'refs/heads/develop' || github.event_name == 'release'
-
     permissions:
-      id-token: write   # This is required for requesting the JWT
-      contents: read    # This is required for actions/checkout
-      packages: write  # This is required for pushing to ghcr
-      attestations: write # This is required for pushing the attestation
-      artifact-metadata: write # This is required for creating the storage record
```
pjonsson, line 39: Does the new workflow not use the GitHub attestation action somewhere under the hood?

This permission seems to be required for getting storage records on the linked artifact page when using the attestation action: https://docs.github.com/en/actions/how-tos/secure-your-work/use-artifact-attestations/use-artifact-attestations#uploading-artifacts-to-the-linked-artifacts-page

.github/workflows/docker.yml
```diff
@@ -26,95 +24,93 @@ concurrency:
...
+            echo "${META_TAGS}"
+            echo "EOF"
+          } >> "$GITHUB_OUTPUT"
+
+  # Build (and push) a combined linux/amd64 + linux/arm64
+  # See https://docs.docker.com/build/ci/github-actions/github-builder/build/
+  # for documentation on this docker maintained workflow.
+  build:
+    needs: prepare
+    uses: docker/github-builder/.github/workflows/build.yml@v1
```
pjonsson, line 84: This should be pinned to a sha256.

.github/workflows/docker.yml
```diff
@@ -26,95 +24,93 @@ concurrency:
...
+        run: |
+          set -euo pipefail
+          echo "meta-images=${REGISTRY}/${IMAGE_NAME}" >> "$GITHUB_OUTPUT"
+
+          # Reproducible-build epoch: timestamp of the last commit that touched
+          # files which affect the built image.
+          SOURCE_DATE_EPOCH=$(git log -1 --pretty=%ct -- pyproject.toml uv.lock cubedash)
+          echo "source-date-epoch=${SOURCE_DATE_EPOCH}" >> "$GITHUB_OUTPUT"
+
+          if [ "${{ github.event_name }}" = "release" ]; then
```
pjonsson, line 58: I only run zizmor in offline mode, but `uvx zizmor -o --persona auditor .github/workflows/docker.yml` complains about this line, and the lack of name (besides the floating `@v1` usage).

Dockerfile
```diff
@@ -1,9 +1,9 @@
 # syntax=docker/dockerfile:1
-FROM ghcr.io/osgeo/gdal:ubuntu-small-3.13.3@sha256:64250faf833c06d4b21afce4c27190039ba7ab58d70f0eebc87cf77d929c0b40 AS base
 
-LABEL org.opencontainers.image.source=https://github.com/opendatacube/datacube-explorer
```
pjonsson, line 4: I genuinely don't understand the difference between labels and annotations, but removing the labels and instead putting annotations into the workflow means that anyone rebuilding the image on their own (like we do, to get security patches) will end up without the metadata.

After a bit of searching, someone in HPC seem to think that the labels should be the ground truth (because you can't misplace them, I think?): https://qnib.org/blog/2023/03/22/deeper-dive-into-container-labels-and-annotations/

pjonsson:
@omad I think you don't get notifications on comments for things that have already been merged, so adding a ping so you don't miss my comments.

pjonsson:
This broke the attestation somehow: https://github.com/opendatacube/datacube-explorer/attestations

The UI isn't great when I have to Google to find where the attestations can be listed.

PR #1239, opened 2026-09-02
[webmaps] Fix Carto Watermarking / Switch basemaps

omad:
## Note

Also includes changes from #1238 !!! 

I wanted to build this as a docker image to test out, and one of the changes there was permitting manual GHA runs to publish dev images.

## Change/Fix

This change switches to using [OpenFreeMap](https://openfreemap.org/) and [OpenMapTiles](https://openmaptiles.org/styles/) for the webmap baselayers, to avoid requiring every deployment of Explorer register and integrate a Carto API key.

This new basemap uses vector map tiles instead of raster tiles, which are newer, faster, lighter, better... But, don't work with native Leaflet JS. So, I've pulled in [maplibre-gl-leaflet](https://github.com/maplibre/maplibre-gl-leaflet) PLUS maplibre-gl as a quick fix.

## CI Breakage

The failure that's breaking all the CI builds in the integration tests is a slight difference in floating point values for one of the fixtures. It should be fixed, but, it's not urgent and should be ignored for now.

## Live Deployment Test

https://explorer.dev.easi-eo.solutions/products/ga_ls_wo_3

image

## Reference broken maps

image

----
📚 Documentation preview 📚: https://datacube-explorer--1239.org.readthedocs.build/en/1239/


SpacemanPaul:
I'm OK with merging this for now I think.

I'm most of the way through a different approach that would make the basemap user configurable, that I still think is a good idea.  (I suspect more and more free basemap suppliers will be requiring api keys over the coming months to years and I don't want this to become a game of whack-a-mole).

SpacemanPaul (approved):

Issue #1092, opened 2026-04-03
Multiple `settings.env.py` required to configure Explorer


pjonsson:
The deployment documentation states:

> Add a file to the current directory called ``settings.env.py``

but the default settings file in the container is `/app/lib/python3.12/site-packages/settings.env.py`, so a user is unlikely to find that.

The default time zone is picked up from the `settings.env.py` in the current working directory instead of `BASE_DIR`, so that is `/home/ubuntu` for the container.

Explorer shouldn't be expecting different settings to be specified in files with the same name, but mounted in different places, and the documentation should be clear on where the settings file is supposed to be.

SpacemanPaul:
Also some configuration settings can be set from the environment, most have to be read from settings.env.py.

PR #1245, opened 2026-09-03
Fully configurable basemaps

SpacemanPaul:
Extension to #1239 

- support both vector and raster base maps.
- make choice of basemap and associated map copyright attribution fully configurable at runtime.
- Upgrade to leaflet 1.9.4 and latest versions of node/npm.
- Documentation and tests.

Basemap URL is set with config option `CUBEDASH_BASEMAP_URL_TEMPLATE` and is treated as a vector basemap by default, but as a raster basemap if the template includes standard `{x}`, `{y}` and `{z}` placeholders.

Attribution is set with config option `CUBEDASH_BASEMAP_ATTRIBUTION` which may contain arbitrary HTML (so can embed `<a href>` tags for links and HTML entities like `&copy;` for a copyright symbol.)

The defaults for these settings are the OpenFreeMap Postitron basemap introduced in #1239, with appropriate attribution.

----
📚 Documentation preview 📚: https://datacube-explorer--1245.org.readthedocs.build/en/1245/


cubedash/static/tsconfig.json
```diff
@@ -62,5 +61,8 @@
     /* Experimental Options */
     // "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
     // "emitDecoratorMetadata": true,         /* Enables experimental support for emitting type metadata for decorators. */
+
+    "allowUmdGlobalAccess": true,
+    "types": ["leaflet/index.d.ts"]
```
pjonsson: I don't see a index.d.ts anywhere and there's no leaflet directory in the repository as far as I can tell, is this a debug leftover?

If you still need this, I'm not sure this is the way to configure tsc for third-party d.ts files. I did a quick search and it seems that there is a consensus that typeRoots is the wrong directive for it, but besides that responses were conflicted. Some responses pointed to using include directive instead.

cubedash/static/overview.ts
```diff
@@ -222,15 +222,28 @@ class DatasetsLayer extends L.GeoJSON {
...
 class OverviewMap extends L.Map {
     constructor(private dataLayers: DataLayer[],
                 activeLayer: DataLayer | null,
                 defaultZoom: number,
                 defaultCenter: L.LatLngTuple) {
-        super("map", {
-            zoom: defaultZoom,
-            center: defaultCenter,
-            layers: [
+        var layers = [];
```
pjonsson: If this has to be re-assignable, use let instead of var.

Isn't this const though? The content of layers is updated, but not the reference to the container itself? (Or am I wrong about how TypeScript uses const?)

cubedash/static/overview.ts
```diff
@@ -222,15 +222,28 @@ class DatasetsLayer extends L.GeoJSON {
...
+        if (basemapIsRaster) {
```
pjonsson: Can the old structure of the code remain if you change it to:

```typescript
layers: [basemapIsRaster ? L.tileLayer(..) : (L as any).maplibreGL(..)],
```
?

cubedash/static/overview.ts
```diff
@@ -222,15 +222,28 @@ class DatasetsLayer extends L.GeoJSON {
     }
 }
 
+declare const basemapTemplateUrl: string;
```
pjonsson, line 226: How do these get injected here?

cubedash/templates/dataset.html
```diff
@@ -183,15 +183,29 @@ <h3 class="followed">
 
     {% if extent_geojson %}
         <script type="application/javascript">
+            const basemapTemplateUrl = "{{ basemap_url_template }}";
+            const basemapAttribution = "{{ basemap_attribution }}";
+            const basemapIsRaster = {% if basemap_is_raster %}true{%  else %}false{% endif %};
+
+            var osm_carto_light = {};
```
pjonsson: I would prefer to keep `osm_carto_light` const like it was before. This part is pure JavaScript so there are no types that can get in the way, won't `const osm_carto_light = basemapIsRaster ? L.tileLayer(..) : L.maplibreGL(..);` work?

If the variable has to be non-const, I think it should be introduced with `let` instead of `var`.

The initialization in the RHS in the current code is redundant, the variable always gets initialized in the following lines.

Edit: on second thought, shouldn't this part of the code be templated instead of templating `basemapIsRaster`?

cubedash/templates/layout/base.html
```diff
@@ -151,8 +151,8 @@
             function updateAttribution(e) {
                 var attributions = Object.keys(map._layers).map(function (name) {
                     var layer = map._layers[name];
-                    if (layer.getAttribution) {
-                        return layer.getAttribution();
+                    if (layer.options && layer.options.attribution) {
+                        return layer.options.attribution;
```
pjonsson, line 159: The previous code returned the result of a method call, this returns a reference to a method?

cubedash/templates/layout/base.html
```diff
@@ -122,7 +122,7 @@
     <div id="quiet-page-errors"></div>
 </footer>
 
-<script src="{{ url_for('static', filename='leaflet-1.7.1/leaflet.js') }}"></script>
+<script src="{{ url_for('static', filename='leaflet-1.9.4/leaflet.js') }}"></script>
 <script src="{{ url_for('static', filename='maplibre-gl-5.24.0/maplibre-gl.js') }}"></script>
 <script src="{{ url_for('static', filename='maplibre-gl-leaflet-0.1.4/leaflet-maplibre-gl.js') }}"></script>
```
pjonsson: The includes for the WebGL libraries should be conditional on whether they are required for the server configuration.

cubedash/_model.py
```diff
@@ -184,6 +196,11 @@ def handle_exception(e: HTTPException) -> tuple[ResponseValue, int]:
     return app
 
 
+def basemap_is_raster(url: str) -> bool:
+    """Basemap URL is raster if it contains x,y and z placeholders."""
```
pjonsson: Missing space after the first comma.

docs/config.rst
```diff
@@ -42,6 +42,31 @@ The following configuration settings are provided by Datacube Explorer:
...
+
+    For vector tile base maps, this is the style URL passed to MapLibre.
+
+    For raster tile base maps, this is the URL template passed to Leaflet, and should
+    contain `{z}`, `{x}` and `{y}` placeholders.  The presence of these template
+    placeholders is used to distinguish between vector and raster tile implementations.
+
+    The default is the positron-style vector basemap from OpenFreeMap. (`https://tiles.openfreemap.org/styles/positron`)
+
+    For raster tile providers that require an API Key, embed the API key in the raster map url template as
```
pjonsson, line 70: Is the variable configured in `settings.env.py`? I don't think we want to encourage people to put their keys in the file, can we put in an example of how to get the API key from an environment variable?

integration_tests/test_configurable_page_elements.py
```diff
@@ -44,6 +44,29 @@ def app_configured_client(client: FlaskClient):
...
             ],
+            "CUBEDASH_BASEMAP_URL_TEMPLATE": "https://vector.example.com/styles/dummy",
+            "CUBEDASH_BASEMAP_ATTRIBUTION": '&copy;<a href="https://example.com/attribution">Vector Basemaps</a>',
+        }
+    )
+    return client
+
+
+@pytest.fixture()
+def app_configured_raster_basemap_client(client: FlaskClient):
```
pjonsson: Missing return type.

cubedash/templates/product.html
```diff
@@ -324,6 +324,14 @@ <h3 class="followed">
     {{ super() }}
 
     {% from "layout/macros.html" import chart_timeline %}
+    <script>
+        const basemapTemplateUrl = "{{ basemap_url_template }}";
+        const basemapAttribution = "{{ basemap_attribution }}";
+        const basemapIsRaster = {% if basemap_is_raster %}true{%  else %}false{% endif %};
+
+        // Not sure why this is needed, but it makes things work.
+        var exports = {}
```
pjonsson, line 333: I'm guessing this  is overwriting the exports for the module (https://stackoverflow.com/questions/21471103/what-is-a-module-and-difference-between-module-exports-vs-exports).

Could it  be caused by the script tag of this block is missing the `type="application/javascript"` part?

Edit: `exports` is a reference to `module.exports` according to this page: http://brunops.org/exports-vs-module-exports

cubedash/_model.py
```diff
@@ -104,6 +104,18 @@ def create_app(test_config=None) -> flask.Flask:
...
         CORS(app, resources=[r"/stac/*", r"/api/*"])
 
+    # Basemap defaults
+    app.config.setdefault(
+        "CUBEDASH_BASEMAP_URL_TEMPLATE", "https://tiles.openfreemap.org/styles/positron"
+    )
+    app.config.setdefault(
+        "CUBEDASH_BASEMAP_ATTRIBUTION",
+        '&copy; <a href="http://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors, '
+        + '&copy; <a href="https://openfreemap.org">OpenFreeMap</a>, '
```
pjonsson, line 114: Not sure why Ruff isn't complaining, but can't we use the implicit string concatenation in Python and omit the `+`?

pjonsson (dismissed):
Having had some time to reflect, setting the exports to the empty dictionary feels like a brittle hack, so I really feel we should fix the underlying problem before merging this and getting it into a release.

cubedash/static/tsconfig.json
SpacemanPaul: Yep, that line can go.

cubedash/static/overview.ts
SpacemanPaul: Neither typescript nor straight javascript are my strong suits. I just went with the first thing that worked.  I'll try a few other approaches.

cubedash/static/overview.ts
SpacemanPaul: That works, although I'm not convinced it's easier to read.

cubedash/static/overview.ts
SpacemanPaul: From the previous script tag in the `product.html` template.

I'll add a comment.

cubedash/templates/dataset.html
SpacemanPaul: Oo nice cleanup.  I was just keeping things as consistent with the overview.ts page as possible, but that makes sense.

cubedash/templates/layout/base.html
SpacemanPaul: The previous code only worked for leaflet raster basemaps.  This works for both raster and vector basemaps.

The behaviour for mapboxlibre.gl vector layers was weird.  When I stepped through the code with the browser webdevtools debugger, I got the attribution from the provider style definition - not the attribution passed in from config to the layer constructor.   And when I turned off all the breakpoints, it didn't work at all - didn't display ANY attribution.   I presume there must be some race condition with maplibre initialisation happening.

This code always returns the attribution passed in, for both basemap types.

docs/config.rst
SpacemanPaul: Given the API key gets placed in the URL in plaintext, and so is viewable in "Network" tag of the client browser webdevtools, and is visible in plaintext in the javascript if you follow the provider's directions, I don't really see the point in trying to hide it in config as a security practice.

Of course there might be technical reasons to shift config to environment variables for some deployments, and there already is a section about this in the config docs ...  although it doesn't seem to work as documented.

SpacemanPaul: I've updated the docs, but generalised configuration by environment variable will need to wait for a future PR.  See my comment on #1092

cubedash/templates/product.html
SpacemanPaul: I'm outside of my comfort zone here and I do not understand exactly what is happening here, but I'll read your links and see what I can figure out.

SpacemanPaul: It's to do with the way the Typescript compiler was configured to import modules.  I'm pretty sure what I had is fine and harmless, but it's cleaner now.

SpacemanPaul:
> Having had some time to reflect, setting the exports to the empty dictionary feels like a brittle hack, so I really feel we should fix the underlying problem before merging this and getting it into a release.

I don't thinks it's as bad as you suggest, but I've got it working without and it at least looks cleaner now.

andrewdhicks (approved):
LGTM

cubedash/templates/product.html
```diff
@@ -324,6 +324,14 @@ <h3 class="followed">
     {{ super() }}
 
     {% from "layout/macros.html" import chart_timeline %}
+    <script type="application/javascript">
+        const basemapTemplateUrl = "{{ basemap_url_template }}";
+        const basemapAttribution = "{{ basemap_attribution }}";
+        const basemapIsRaster = {% if basemap_is_raster %}true{%  else %}false{% endif %};
+
+        // Not sure why this is needed, but it makes things work.
+        var exports = {}
```
pjonsson, line 333: @SpacemanPaul this part was still merged to develop, did you forget to commit the fix to this before pushing?

PR #1253, opened 2026-09-10
Dockerfile: reduce gunicorn timeout

pjonsson:
Remove the timeout argument to gunicorn
so workers get killed after the default
value of 30 seconds of silence.

The people who have so large databases
that they need a larger timeout
are most likely already deploying
Explorer in custom ways, so this
change shouldn't affect them.

----
📚 Documentation preview 📚: https://datacube-explorer--1253.org.readthedocs.build/en/1253/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -127,8 +127,6 @@ CMD ["gunicorn", \
      "--threads=2", \
      "-k", \
      "gthread", \
-     "--timeout", \
-     "90", \
      "--config", \
      "python:cubedash.gunicorn_config", \
      "cubedash.startup_utils:create_app()"]


omad (approved):

omad:
@pjonsson 
I can't see a problem with this. I am curious about the reason for it though.

pjonsson:
> I can't see a problem with this. I am curious about the reason for it though.

Why it was 90 seconds in the first place, or why I'm lowering the timeout? I don't know the history for the value, but if an Explorer instance gets indexed by AI crawlers it will eventually (or fairly quickly) stop responding to most other requests because everything gums up, so this makes Gunicorn step in and kill non-responsive parts faster.

PR #1260, opened 2026-09-14
Minify overview.js

pjonsson:
Use tsc for typechecking and
use SWC for minification.
This saves ~4k on overview.js.

----
📚 Documentation preview 📚: https://datacube-explorer--1260.org.readthedocs.build/en/1260/


pjonsson:
I'm not sure how I would detect that something breaks with overview.js, but the product page looks fine except for the grey Canvas (which is caused by my lack of WebGL) when I test this locally.

